Issue #6, opened 2019-08-01
Do not delete the kcov_output in profile devel


andreeaflorescu:
When running the integration tests with profile devel, we can leave the kcov_output directory so that it is easier to inspect the coverage report.
Alternatively, we can add a `--no-cleanup` so we don't leave lingering files on the system without the user being aware.

Issue #10, opened 2019-12-03
rust-vmm ci coverage test on arm64


MrXinWang:
As kcov seems to work on arm64, I think it is the time to consider fully enable and test the ci coverage test of rust-vmm on arm64. @andreeaflorescu @michael2012z

FYI, please check the attached `kcov_output` dir for `kvm-ioctls`
[kcov_output.zip](https://github.com/rust-vmm/rust-vmm-ci/files/3914404/kcov_output.zip)

MrXinWang:
Closing this issue as PR #11 has been merged.

PR #11, opened 2019-12-04
Enable rust-vmm coverage test in CI

MrXinWang:
Referencing discussion in [kvm-ioctls/issue#9](https://github.com/rust-vmm/kvm-ioctls/issues/9) and [issue#10](https://github.com/rust-vmm/rust-vmm-ci/issues/10), proposing a draft PR for enabling coverage test on arm64.

I think we also need to rename the old coverage path file in each repository to avoid misunderstanding. What do you think @andreeaflorescu ? 

Signed-off-by: Henry Wang <<henry.wang@arm.com>>


MrXinWang:
Updated the commit to fix #6

andreeaflorescu:
@MrXinWang I published a new version of rust-vmm-container (v5). I will check how your PR works with kvm-ioctls and I will dismiss my review so we can merge this one.

MrXinWang:
@andreeaflorescu Thanks very much for that. I remember from our last discussion we also need to upload the coverage.json file for aarch64 to seperate the x86_64 and arm64 coverage files.

integration_tests/test_coverage.py
```diff
@@ -87,7 +89,6 @@ def _get_current_coverage(coverage_config):
         )[0])
 
     # Remove coverage related directories.
-    shutil.rmtree(kcov_ouput_dir, ignore_errors=True)
```
andreeaflorescu: Why are we not removing this anymore?

MrXinWang: According to my understanding, this line of code removes the kcov_output directory, which contains the coverage information (results) generated by kcov. In your posted issue #6 , it is asked to not remove this directory. I think probably we can fix that issue here...

MrXinWang:
@andreeaflorescu yes sure. I will upload a commit together with the coverage file thing after you are happy with every change in this PR :)

integration_tests/test_coverage.py
andreeaflorescu: Let's have a separate commit for that. We want to keep it only for `profile --devel`. Here we are keeping it on every profile which is not the desired behavior because when calling this via the CI we do want to clean up all the files before exiting the test. In `profile --devel` this would be useful for contributors to see what lines are not covered.

andreeaflorescu:
> @andreeaflorescu yes sure. I will upload a commit together with the coverage file thing after you are happy with every change in this PR :)

To recap, to merge this we need to do the following:
- create 2 example coverage files that should replace the current `coverage_config.json.sample`. This should be named `coverage_config_x86_64.json.sample` and `coverage_config_aarch64.json.sample` so that people can just remove the `sample` suffix and use them as an example.
- Update the README.md to specify that we now have 2 coverage files, one per each platform. The section that needs to be updated is `## Getting Started with rust-vmm-ci`, bullet 2.
 - update the Buildkite pipeline to use rust-vmm/dev:v5

I hope I am not missing anything. After these changes are address, we can merge the PR.

integration_tests/test_coverage.py
MrXinWang: Sure. I have reverted the code. Thanks!

MrXinWang:
@andreeaflorescu Thanks for the recap. I think I have changed all of them, also changed the container version used in README.md from v3 to v5. Please review.

MrXinWang:
@andreeaflorescu I am sorry but do we both missed the one of the most important thing: add another pipeline in Buildkite for arm coverage test...? I have updated the code, now it looks ok to me.

```
- label: "coverage-arm"
    commands:
      - pytest rust-vmm-ci/integration_tests/test_coverage.py
    retry:
      automatic: false
    agents:
      platform: arm.metal
      os: linux
    plugins:
      - docker#v3.0.1:
          privileged: true
          image: "rustvmm/dev:v5"
          always-pull: true
```

andreeaflorescu:
@MrXinWang hahaha yap, we totally missed that.

Issue #9, opened 2019-11-02
Add test to check signed-off-by


andreeaflorescu:
A recurrent problem we have in rust-vmm PRs is that we have unsigned commits. Let's make this part of the CI as it is easy to miss during reviews.

Issue #14, opened 2020-01-15
Add test to check the commit message format


andreeaflorescu:
Add a test to check that the commit message has an appropriate format:
- it has a summary no longer than 50 chars (we can also adjust this number if we think it's too small)
- each line of the body is less than 72 chars

andreeaflorescu:
Fixed by #27

PR #27, opened 2020-07-21
Add test for commit message format

lauralt:
Fixes: #9, #14.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -190,3 +190,16 @@ steps:
           privileged: true
           image: "rustvmm/dev:v6"
           always-pull: true
+
+  - label: "commit-format"
+    commands:
+      - pytest rust-vmm-ci/integration_tests/test_commit_format.py
+    retry:
+      automatic: false
+    agents:
+      os: linux
+    plugins:
+      - docker#v3.0.1:
+          image: "rustvmm/dev:v6"
+          always-pull: true
+          propagate-environment: true
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -11,7 +11,7 @@ steps:
       os: linux
     plugins:
       - docker#v3.0.1:
-          image: "rustvmm/dev:v5"
+          image: "rustvmm/dev:v6"
           always-pull: true
 
   - label: "build-gnu-arm"
@@ -24,7 +24,7 @@ steps:
       os: linux
     plugins:
       - docker#v3.0.1:
-          image: "rustvmm/dev:v5"
+          image: "rustvmm/dev:v6"
           always-pull: true
 
   - label: "build-musl-x86"
@@ -37,7 +37,7 @@ steps:
       os: linux
     plugins:
       - docker#v3.0.1:
-          image: "rustvmm/dev:v5"
+          image: "rustvmm/dev:v6"
           always-pull: true
 
   - label: "build-musl-arm"
@@ -50,7 +50,7 @@ steps:
       os: linux
     plugins:
       - docker#v3.0.1:
-          image: "rustvmm/dev:v5"
+          image: "rustvmm/dev:v6"
           always-pull: true
 
   - label: "style"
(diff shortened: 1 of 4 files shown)


.buildkite/pipeline.yml
```diff
@@ -190,3 +190,17 @@ steps:
...
+    commands:
+      - pytest rust-vmm-ci/integration_tests/test_commit_format.py
+    retry:
+      automatic: false
+    agents:
+      platform: x86_64.metal
+      os: linux
+    plugins:
+      - docker#v3.0.1:
+          privileged: true
```
aghecenco, line 204: Does the container need root privileges for this test?

integration_tests/test_commit_format.py
```diff
@@ -0,0 +1,53 @@
...
+def test_commit_format():
+    """
+    Checks commit message format for the current PR's commits.
+
+    Checks if commit messages follow the 50/72 git commit rule
+    and if commits are signed.
+    """
+    # Get hashes of PR's commits in their abbreviated form for
+    # a prettier printing.
+    shas_cmd = "git log --no-merges --pretty=%h --no-decorate " \
```
aghecenco, line 23: `git` won't be available in the container until v6 is published

integration_tests/test_commit_format.py
```diff
@@ -0,0 +1,53 @@
...
+COMMIT_TITLE_MAX_LEN = 50
+COMMIT_BODY_LINE_MAX_LEN = 72
+BASE_BRANCH = os.environ['BUILDKITE_PULL_REQUEST_BASE_BRANCH']
+
+
+def test_commit_format():
+    """
+    Checks commit message format for the current PR's commits.
+
+    Checks if commit messages follow the 50/72 git commit rule
```
aghecenco, line 18: nit: after the explanation, add a link to the 50/72 rule so it doesn't look like they're just arbitrary numbers we happen to like

integration_tests/utils.py
```diff
@@ -9,3 +11,12 @@ def get_repo_root_path():
     rust_vmm_ci_path = os.path.dirname(integration_tests_path)
 
     return os.path.dirname(rust_vmm_ci_path)
+
+
+def get_cmd_output(cmd):
+    """Returns stdout content of `cmd` command."""
+    cmd_out = subprocess.run(cmd, shell=True, check=True,
+                             stdout=subprocess.PIPE,
+                             stderr=subprocess.PIPE)
```
aghecenco, line 20: This function doesn't use `stderr` for anything, so no need to pipe it

.buildkite/pipeline.yml
lauralt: Nope, it shouldn't.

integration_tests/test_commit_format.py
lauralt: Wanted to, then I forgot :(, thanks.

integration_tests/utils.py
lauralt: Done.

integration_tests/test_commit_format.py
lauralt: Sorry, I was under the impression it was updated in the last PR from there.

aghecenco: The container image still has to be built and pushed to the Docker registry (sounds like something we could automate :thinking: but later); in the meantime I did that and v6 should be available now

.buildkite/pipeline.yml
```diff
@@ -188,5 +188,17 @@ steps:
...
+  - label: "commit-format"
+    commands:
+      - pytest rust-vmm-ci/integration_tests/test_commit_format.py
+    retry:
+      automatic: false
+    agents:
+      os: linux
+    plugins:
+      - docker#v3.0.1:
+          image: "rustvmm/dev:v6"
```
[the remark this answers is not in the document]
andreeaflorescu: https://github.com/buildkite-plugins/docker-buildkite-plugin

lauralt:
It seems the dummy PRs from kvm-ioctls: https://github.com/rust-vmm/kvm-ioctls/pull/105 and https://github.com/rust-vmm/kvm-ioctls/pull/107 are failing right.

PR #28, opened 2020-07-27
Fix commit format test

lauralt:
Even though `origin` was pointing to the right upstream repo, the repo wasn't fetched and because of that, we ended comparing HEAD with a wrong master. The commit format test for the following dummy PR: https://github.com/rust-vmm/kvm-ioctls/pull/109 is passing now (which is the expected behavior).

Also, it looks like we had the same problem in the benchmark test. I fixed that too, but this change (and basically the entire test) can be properly tested in a PR that actually uses it (for example https://github.com/rust-vmm/linux-loader/pull/46). Nevertheless, I expect fetching is necessary here too.

### DIFF
--- a/integration_tests/test_commit_format.py
+++ b/integration_tests/test_commit_format.py
@@ -3,12 +3,14 @@
 """Test the commit message format."""
 
 import os
+import subprocess
 
 from utils import get_cmd_output
 
 COMMIT_TITLE_MAX_LEN = 50
 COMMIT_BODY_LINE_MAX_LEN = 72
 BASE_BRANCH = os.environ['BUILDKITE_PULL_REQUEST_BASE_BRANCH']
+BASE_REPO = os.environ['BUILDKITE_REPO']
 
 
 def test_commit_format():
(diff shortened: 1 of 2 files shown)


integration_tests/test_commit_format.py
```diff
@@ -3,12 +3,16 @@
...
+import subprocess
 
 from utils import get_cmd_output
 
 COMMIT_TITLE_MAX_LEN = 50
 COMMIT_BODY_LINE_MAX_LEN = 72
 BASE_BRANCH = os.environ['BUILDKITE_PULL_REQUEST_BASE_BRANCH']
+# When preparing the test environment, `origin` is set to the
+# current upstream repository ('BUILDKITE_REPO').
+BASE_REPO = "origin"
```
andreeaflorescu, line 15: I would prefer us to use BUILDKITE_REPO here instead of origin. `origin` is just the name, and the name can always be updated. Using the env variable sounds less prone to error to me.

lauralt: I agree, for this test seems cleaner this way. In the benchmark test on the other hand, it seems we are hardcoding `origin` (and other things too) for a nicer `git checkout` command. After `git fetch some_url master`, I think `git checkout FETCH_HEAD` could be used, but anyway, more changes to that test are required and I think the only good way to test them is in a PR that adds some benchmarks or at least a benchmark pipeline.
In the meantime, I updated the commit test.

PR #31, opened 2020-07-29
Fix `test_bench`

andreeaflorescu:
Description in commit message.

### DIFF
--- a/integration_tests/test_benchmark.py
+++ b/integration_tests/test_benchmark.py
@@ -23,18 +23,14 @@ def test_bench():
     # For this purpose, we need to explicitly check the return code.
     assert return_code == 0, "stdout: {}\n stderr: {}".format(stdout, stderr)
 
-    # Move to upstream tip.
-    subprocess.run(
-        "git fetch {} {}".format(REMOTE, BRANCH),
-        shell=True, check=True
-    )
-    subprocess.run(
-        "git checkout {}/{}".format(REMOTE, BRANCH),
-        shell=True, check=True
-    )
-
     # Get numbers from upstream tip, without the changes from the current PR.
+    _checkout_master_branch()
     return_code, stdout, stderr = _run_cargo_bench("before")
+
+    # Before checking any results, let's just go back to the PR branch.
+    # This way we make sure that the cleanup always happens even if the test fails.
+    _checkout_pr_branch()
+
     if return_code == 0:
         # In case this benchmark also ran successfully, we can call critcmp and compare the results.
         _run_critcmp()
@@ -71,3 +67,21 @@ def _run_critcmp():
     print(p.stdout.decode('utf-8'))
     print('ERRORS')
     print(p.stderr.decode('utf-8'))
+
+def _checkout_master_branch():
+    subprocess.run(
+        "git fetch {} {}".format(REMOTE, BRANCH),
+        shell=True, check=True
+    )
+    subprocess.run(
+        "git checkout {}/{}".format(REMOTE, BRANCH),
+        shell=True, check=True
+    )
+
+def _checkout_pr_branch():
+    subprocess.run(
+        "git checkout -",
+        shell=True, check=True,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE
+    )
--- a/integration_tests/test_benchmark.py
+++ b/integration_tests/test_benchmark.py
@@ -18,7 +18,10 @@ def test_bench():
     os.chdir(get_repo_root_path())
 
     # Get numbers for current HEAD.
-    _run_cargo_bench("after")
+    return_code, stdout, stderr = _run_cargo_bench("after")
+    # Even if it is the first time this test is run, the benchmark tests should pass.
+    # For this purpose, we need to explicitly check the return code.
+    assert return_code == 0, "stdout: {}\n stderr: {}".format(stdout, stderr)
 
     # Move to upstream tip.
     subprocess.run(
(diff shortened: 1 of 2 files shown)


andreeaflorescu:
Happy case for when master does not have performance tests is tested here: https://github.com/rust-vmm/event-manager/pull/31

integration_tests/test_benchmark.py
```diff
@@ -26,18 +26,14 @@ def test_bench():
...
-    subprocess.run(
-        "git checkout {}/{}".format(REMOTE, BRANCH),
-        shell=True, check=True
-    )
-
     # Get numbers from upstream tip, without the changes from the current PR.
+    _checkout_master_branch()
     return_code, stdout, stderr = _run_cargo_bench("before")
+
+    # Before checking any results, let's just go back to the PR branch.
```
alexandruag, line 33: Is is guaranteed that tests won't run in parallel for the same source folder?

integration_tests/test_benchmark.py
```diff
@@ -18,22 +18,50 @@ def test_bench():
...
 
     # Get numbers for current HEAD.
-    _run_cargo_bench("after")
+    return_code, stdout, stderr = _run_cargo_bench("after")
+    # Even if it is the first time this test is run, the benchmark tests should pass.
+    # For this purpose, we need to explicitly check the return code.
+    if return_code != 0:
+        print(stdout)
+        print(stderr)
+        assert return_code == 0
```
lauralt, line 27: I think this could be replaced with something like this:
```python
assert return_code == 0, "stdout: {}, stderr: {}".format(stdout, stderr)
```

integration_tests/test_benchmark.py
```diff
@@ -18,22 +18,50 @@ def test_bench():
...
+        # The benchmark did not run successfully, but it might be that it is because a benchmark does not exist.
+        # In this case, we do not want to fail the test.
+        if "error: no bench target named `main`" in stderr:
+            # This is a bit of a &*%^ way of checking if the benchmark does not exist.
+            # Hopefully it will be possible to check it in another way...soon
+            print("There are no benchmarks in master. No comparison can happen.")
+        else:
+            print(stdout)
+            print(stderr)
+            assert return_code == 0
```
lauralt, line 50: Also, here if we are not interested in printing the return_code since we also print stdout, stderr, we could replace it with: `assert False, "some message"` or just `assert False`.

integration_tests/test_benchmark.py
```diff
@@ -18,22 +18,50 @@ def test_bench():
...
+
+
+def _run_cargo_bench(baseline):
+    """Runs `cargo bench` and tags the baseline."""
+    process = subprocess.run(
```
lauralt, line 55: nit maybe rename `process` with `p` (same as in `_run_critcmp()`) or `proc`, this way the return line would be somehow shorter.

integration_tests/test_benchmark.py
```diff
@@ -18,22 +18,50 @@ def test_bench():
...
 
-    # Get numbers from upstream tip, without the changes from the current PR.
-    _run_cargo_bench("before")
+    return process.returncode, process.stdout.decode('utf-8'), process.stderr.decode('utf-8')
 
-    # Compare.
+
+def _run_critcmp():
     p = subprocess.run(
         "critcmp before after",
```
lauralt, line 66: nit: `before` and `after` could be defined as constants.

integration_tests/test_benchmark.py
```diff
@@ -76,3 +72,21 @@ def _run_critcmp():
...
     print('ERRORS')
     print(p.stderr.decode('utf-8'))
+
+def _checkout_master_branch():
+    subprocess.run(
+        "git fetch {} {}".format(REMOTE, BRANCH),
+        shell=True, check=True
+    )
+    subprocess.run(
+        "git checkout {}/{}".format(REMOTE, BRANCH),
```
alexandruag, line 82: nit: I think we could do a `git checkout FETCH_HEAD` here instead.

integration_tests/test_benchmark.py
alexandruag: Or `raise Exception('message')` I think, but I don't know what the best practice is here.

integration_tests/test_benchmark.py
andreeaflorescu: There cannot be two tests running on the same directory at the same time. One agent can run maximum one test at the time. If there are multiple agents running tests, than each of them will have their own directory in which to check out code.

integration_tests/test_benchmark.py
```diff
@@ -8,28 +8,31 @@
...
-REMOTE = "origin"
-# Possibly prone to changes.
-BRANCH = "master"
-
+GIT_REMOTE_NAME = "origin"
+# Base branch as configured with Buildkite.
+BASE_BRANCH = os.environ['BUILDKITE_PULL_REQUEST_BASE_BRANCH']
+# File used for saving the results of cargo bench when running on the PR branch.
+PR_BENCH_RESULTS_FILE = "pr_bench_results"
+# File used for saving the results of cargo bench when running on the PR branch.
```
lauralt, line 16:
```suggestion
# File used for saving the results of cargo bench when running on the upstream branch.
```

integration_tests/test_benchmark.py
```diff
@@ -68,17 +71,17 @@ def _run_critcmp():
     print('ERRORS')
     print(p.stderr.decode('utf-8'))
 
-def _checkout_master_branch():
+def _git_checkout_upstream_branch():
     subprocess.run(
-        "git fetch {} {}".format(REMOTE, BRANCH),
+        "git fetch {} {}".format(GIT_REMOTE_NAME, BASE_BRANCH),
```
lauralt, line 76: nit: from what it looks like, fetching works with links too, so GIT_REMOTE_NAME could be 'BUILDKITE_REPO' here too.

andreeaflorescu: Good idea. Will do.

integration_tests/test_benchmark.py
andreeaflorescu: I kept the assert, but made it shorter as per @lauralt's previous suggestion.

Issue #8, opened 2019-10-22
Add test for cargo audit


andreeaflorescu:
We shouldn't have too many crates that we depend on, but still it is worth checking cargo audit as part of our CI.

Detail here: https://github.com/RustSec/cargo-audit

anth1y:
@andreeaflorescu I'd like to be assigned to this issue.. but i think i need to be added to the org ?

andreeaflorescu:
Hey @anth1y, I've assign you. I also thought that you need to be part of an org to be assigned to an issue, but it seems that has changed.

Let me know if you need help with anything :)

PR #36, opened 2020-09-06
Create cargo-audit test to scan for vuln

anth1y:
fixes  https://github.com/rust-vmm/rust-vmm-ci/issues/8

cc/ @andreeaflorescu

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -203,3 +203,16 @@ steps:
           image: "rustvmm/dev:v6"
           always-pull: true
           propagate-environment: true
+
+  - label: "cargo-audit"
+    commands:
+      - cargo audit
+    retry:
+      automatic: false
+    agents:
+      os: linux
+    plugins:
+      - docker#v3.0.1:
+          image: "rustvmm/dev:v6"
+          always-pull: true
+          propagate-environment: true
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ steps:
 ```
 
 To see all steps in the pipeline check the
-[.buikite/pipeline.yml](.buildkite/pipeline.yml) file.
+[.buildkite/pipeline.yml](.buildkite/pipeline.yml) file.
 
 ### Custom Pipeline
 
@@ -156,7 +156,7 @@ being downloaded.
 ## Integration Tests
 
 In addition to the one-liner tests defined in the
-[Builkite Pipeline](#buildkite-pipeline), the rust-vmm-ci also has more
+[Buildkite Pipeline](#buildkite-pipeline), the rust-vmm-ci also has more
 complex tests defined in [integration_tests](integration_tests).
 
 ### Test Profiles
--- a/README.md
+++ b/README.md
@@ -238,7 +238,18 @@ steps:
 ```
 
 The test requires [`criterion`](https://github.com/bheisler/criterion.rs)
-benchmarks to be exported by the crate. `criterion` collects performance
+benchmarks to be exported by the crate. The test expects the entry point
+into the performance benchmarks to be named `main`. In other words, the
+following configuration is expected in `Cargo.toml`:
+```toml
+[[bench]]
+name = "main"
+```
+
+All benchmarks need to be collected in a main.rs file placed in `benches/`.
+
+
+`criterion` collects performance
 results by running a function for a user-configured number of iterations,
 timing the runs, and applying statistics. The individual benchmark tests must
 be added in the crate. They can be run outside the CI with:
--- a/integration_tests/test_benchmark.py
+++ b/integration_tests/test_benchmark.py
@@ -23,18 +23,14 @@ def test_bench():
     # For this purpose, we need to explicitly check the return code.
     assert return_code == 0, "stdout: {}\n stderr: {}".format(stdout, stderr)
 
-    # Move to upstream tip.
-    subprocess.run(
-        "git fetch {} {}".format(REMOTE, BRANCH),
-        shell=True, check=True
-    )
-    subprocess.run(
-        "git checkout {}/{}".format(REMOTE, BRANCH),
-        shell=True, check=True
-    )
-
     # Get numbers from upstream tip, without the changes from the current PR.
+    _checkout_master_branch()
     return_code, stdout, stderr = _run_cargo_bench("before")
+
+    # Before checking any results, let's just go back to the PR branch.
+    # This way we make sure that the cleanup always happens even if the test fails.
+    _checkout_pr_branch()
+
     if return_code == 0:
         # In case this benchmark also ran successfully, we can call critcmp and compare the results.
         _run_critcmp()
@@ -71,3 +67,21 @@ def _run_critcmp():
     print(p.stdout.decode('utf-8'))
     print('ERRORS')
     print(p.stderr.decode('utf-8'))
+
+def _checkout_master_branch():
+    subprocess.run(
+        "git fetch {} {}".format(REMOTE, BRANCH),
+        shell=True, check=True
+    )
+    subprocess.run(
+        "git checkout {}/{}".format(REMOTE, BRANCH),
+        shell=True, check=True
+    )
+
+def _checkout_pr_branch():
+    subprocess.run(
+        "git checkout -",
+        shell=True, check=True,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE
+    )
--- a/integration_tests/test_benchmark.py
+++ b/integration_tests/test_benchmark.py
@@ -18,7 +18,10 @@ def test_bench():
     os.chdir(get_repo_root_path())
 
     # Get numbers for current HEAD.
-    _run_cargo_bench("after")
+    return_code, stdout, stderr = _run_cargo_bench("after")
+    # Even if it is the first time this test is run, the benchmark tests should pass.
+    # For this purpose, we need to explicitly check the return code.
+    assert return_code == 0, "stdout: {}\n stderr: {}".format(stdout, stderr)
 
     # Move to upstream tip.
     subprocess.run(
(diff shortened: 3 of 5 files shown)


integration_tests/test_cargo_audit.py
```diff
@@ -0,0 +1,10 @@
+import os
+
+# cargo-audit reads and checks for the presence of Cargo.lock
```
andreeaflorescu, line 3: How about instead of checking for the existence of the `Cargo.lock` file, we just do a `cargo build --all-features --release` before and then call `cargo audit`?

One of the tests of the rust-vmm-ci is checking the build anyway, so we will end up always in 2 situations:
- the build command was already run and `cargo build` will be a no-op.
- the build command was not run, and the build will happen as part of this command; when the build command will be run next time, it will be a no-op.

Running `cargo build` and then `cargo audit` should not incur a significant delay.

andreeaflorescu:
The script should also be called from the Buildkite pipeline. You can take the `commit-format` test as an example: https://github.com/rust-vmm/rust-vmm-ci/blob/e58ea7445ace0cb984f8002ba2436c34cf592efe/.buildkite/pipeline.yml#L194

anth1y:
@andreeaflorescu  is there a way I can iterate through Buildkite locally so I can see what's going on ? like directory structure etc... Also sorry it's taken me so long

andreeaflorescu:
@anth1y no problem. We don't really recreate the Buildkite environment locally (don't know if it is even possible because it needs to communicate with the main agent). Why do you need to recreate the buildkite env?

You can test the script as follows (and this is what we've done for similar scripts - it's ugly, but it works):
- create a fork of a rust-vmm repository with code; you can use [kvm-ioctls](https://github.com/rust-vmm/kvm-ioctls) for this
- update the rust-vmm-ci to point to your fork instead of the upstream; @lauralt also did that to see if `test_commit_message` works as expected
- once the rust-vmm-ci points to your fork, you can run the test by running the following commands in the `kvm-ioctls` directory:
```
docker run -it  --volume $(pwd):/kvm-ioctls rustvmm/dev:v6
# in the container now
cd kvm-ioctls
python rust-vmm-ci/integration_tests/test_cargo_audit.py
```
- you can now also test it with buildkite by opening a PR with the changes you've done locally to the rust-vmm-ci; check this one for an example: https://github.com/rust-vmm/kvm-ioctls/pull/109

When I added the label with good first issue I didn't really thought about the fact that testing this is so messy. I'm sorry for that.
If you need any help (like for example with submodule as things there are also messy), let me know. I am also available for chatting on the rust-vmm Slack.

anth1y:
```
root@0126a0975267:/kvm-ioctls/kvm-ioctls# pytest ../rust-vmm-ci/integration_tests/test_cargo_audit.py
============================================ test session starts ============================================
platform linux -- Python 3.6.9, pytest-5.4.3, py-1.9.0, pluggy-0.13.1
rootdir: /kvm-ioctls
collected 1 item

../rust-vmm-ci/integration_tests/test_cargo_audit.py .                                                [100%]

============================================ 1 passed in 28.02s =============================================
```

integration_tests/test_cargo_audit.py
```diff
@@ -0,0 +1,12 @@
+from subprocess import Popen, PIPE, STDOUT
+
+
+def test_cargo_audit():
+
+    cmd1 = 'cargo build --all-features --release'
+    cmd2 = 'cargo-audit audit'
+    prs = Popen("{}; {}".format(cmd1, cmd2), shell=True, stdin=PIPE,
+            stdout=PIPE, stderr=STDOUT, close_fds=True)
+    stdout, nothing = prs.communicate()
```
sameo, line 10: Thanks for the PR @anth1y . Any reason why we can't just run `cargo audit` on the repo?

andreeaflorescu: I think I am the culprit :)) because of this: https://github.com/rust-vmm/rust-vmm-ci/pull/36#discussion_r484308058

I was running `cargo audit` at some point and it would fail if `Cargo.lock` was not in the repository. I doubled checked now and it looks like if Cargo.lock does not exist, it will be created when running `cargo audit`, so all good. We do not need to run `cargo build` before. Good catch, @sameo!

integration_tests/test_cargo_audit.py
```diff
@@ -2,10 +2,8 @@
 
 
 def test_cargo_audit():
-
-    cmd1 = 'cargo build --all-features --release'
-    cmd2 = 'cargo-audit audit'
-    prs = Popen("{}; {}".format(cmd1, cmd2), shell=True, stdin=PIPE,
+    cmd1 = 'cargo-audit audit'
```
andreeaflorescu, line 5: Hmm, can it be named just `cmd`?

andreeaflorescu:
@anth1y I tested the PR with a dummy PR in kvm-ioctls. Looks like it is working: https://buildkite.com/rust-vmm/kvm-ioctls-ci/builds/536#2eacab97-a876-4f46-a4e5-33ee37ffc922

I would also like to see if I introduce a vulnerable dependency if it fails. Will do that tomorrow.
Can you squash your commits and add a single signed commit (containing the Signed-off by: line) with all the changes? Once we validate that the test fails when there are vulnerable dependencies we can merge it.

integration_tests/test_cargo_audit.py
```diff
@@ -0,0 +1,10 @@
+from subprocess import Popen, PIPE, STDOUT
+
+
+def test_cargo_audit():
+    cmd = 'cargo-audit audit'
+    prs = Popen("{}".format(cmd), shell=True, stdin=PIPE,
+            stdout=PIPE, stderr=STDOUT, close_fds=True)
+    stdout, nothing = prs.communicate()
```
andreeaflorescu, line 8: Can we use `subprocess.run` instead? You can take a look at an example here: https://github.com/rust-vmm/rust-vmm-ci/blob/9dd386c2e6d35aabdb2064f3ca0d17f2ffd26d4b/integration_tests/test_commit_format.py#L26

We need to add `check = True` otherwise when there are vulnerable dependencies the test pass even though it shouldn't.

.buildkite/pipeline.yml
```diff
@@ -203,3 +203,16 @@ steps:
           image: "rustvmm/dev:v6"
           always-pull: true
           propagate-environment: true
+
+  - label: "cargo-audit"
+    commands:
+      - pytest rust-vmm-ci/integration_tests/test_cargo_audit.py
```
andreeaflorescu, line 209: We should run this with `-s` so that we can see what are the vulnerable crates in case there are any:
```pytest rust-vmm-ci/integration_tests/test_cargo_audit.py -s```

File: integration_tests/test_cargo_audit.py [hunk 6]
```diff
@@ -0,0 +1,10 @@
...
+    print(stdout)
```
andreeaflorescu, line 9: No need to print the `stdout`.

integration_tests/test_cargo_audit.py
```diff
@@ -0,0 +1,10 @@
+from subprocess import Popen, PIPE, STDOUT
+
+
+def test_cargo_audit():
+    cmd = 'cargo-audit audit'
```
andreeaflorescu, line 5: Running the command as `cargo audit` should be enough.

andreeaflorescu (changes requested):
The test does not fail when there are vulnerable dependencies added. Comments inline.

The summary of your commit does not need to be changed, but can you please update the description of your commit and instead of the existing text to write something like:

```Fixes: https://github.com/rust-vmm/rust-vmm-ci/issues/8```

This helps us keeping track of issues that are fixed. When the PR is merged, the issue will also be automatically closed.

File: integration_tests/test_cargo_audit.py [hunk 8]
```diff
@@ -0,0 +1,10 @@
+from subprocess import Popen, PIPE, STDOUT
```
lauralt, line 1: Please add the license here:

`# SPDX-License-Identifier: Apache-2.0`

integration_tests/test_cargo_audit.py
```diff
@@ -0,0 +1,8 @@
+# SPDX-License-Identifier: Apache-2.0
+from subprocess import Popen, PIPE, STDOUT
```
andreeaflorescu, line 2: You don't need to import Popen, PIPE, STDOUT anymore, you only need `import subprocess`.

integration_tests/test_cargo_audit.py
```diff
@@ -0,0 +1,7 @@
+# SPDX-License-Identifier: Apache-2.0
+import subprocess 
+
+def test_cargo_audit():
+    cmd = 'cargo-audit'
+    prs = subprocess.run("{}".format(cmd), shell=True, check=True)
+    stdout, nothing = prs.communicate()
```
andreeaflorescu, line 7: This line is no longer needed. The command fails with:
```E       AttributeError: 'CompletedProcess' object has no attribute 'communicate'
```

lauralt, line 7: Is this needed?

integration_tests/test_cargo_audit.py
```diff
@@ -0,0 +1,7 @@
+# SPDX-License-Identifier: Apache-2.0
+import subprocess 
+
+def test_cargo_audit():
+    cmd = 'cargo-audit'
```
andreeaflorescu, line 5: The command should be `cargo audit`.

.buildkite/pipeline.yml
```diff
@@ -203,3 +203,16 @@ steps:
           image: "rustvmm/dev:v6"
           always-pull: true
           propagate-environment: true
+
+  - label: "cargo-audit"
+    commands:
+      - pytest rust-vmm-ci/integration_tests/test_cargo_audit.py -s
```
andreeaflorescu, line 209: On a second thought, we can just write `cargo audit` here and it should work :-?
This PR was a :roller_coaster:

integration_tests/test_cargo_audit.py
anth1y: should we just delete this ?

andreeaflorescu: Yes, I think it makes more sense. What do  you say?

lauralt: Here, we can also pass `-q` flag since we're not interested in the output if the test is passing.

anth1y: you should refresh the page this file should have been deleted

andreeaflorescu:
@anth1y the test works as expected :fireworks:
Here is a link to the output: https://buildkite.com/rust-vmm/kvm-ioctls-ci/builds/537#1eea9a80-28d3-49a9-bce6-585ea0ed0bff

So now we are ready to merge it. Can you squash the commits and update the description as per my previous comment: https://github.com/rust-vmm/rust-vmm-ci/pull/36#pullrequestreview-499439731

anth1y:
I've already  updated the description ... and I'll squash the commits tonight on my personal machine ... I'm on my work machine ;)

integration_tests/test_cargo_audit.py
lauralt: It applies to the buildkite fast path too :D.

andreeaflorescu:
@anth1y what I meant is to update the commit message, not the PR description (but updating the PR description is also neat!). When you have the time to squash the commits, can you also make the description of the commit: ```Fixes: https://github.com/rust-vmm/rust-vmm-ci/issues/8```? Thanks for the work!

anth1y:
@andreeaflorescu this is really messy can we just close this PR and I'll open an new PR that's a one liner?

andreeaflorescu:
@anth1y sure, no problem. Closing this one, feel free to open a new one with a one liner.

PR #37, opened 2020-09-09
Coverage test: keep stdin open

aghecenco:
Pytest closes stdin by default, so any processes spawned from it won't have it. As some tests in various rust-vmm crates might need stdin to be open, we mock it with a (currently empty) byte stream, which we can extend in the future.

### DIFF
--- a/integration_tests/test_coverage.py
+++ b/integration_tests/test_coverage.py
@@ -79,7 +79,9 @@ def _get_current_coverage(coverage_config):
         exclude_pattern
     )
 
-    subprocess.run(kcov_cmd, shell=True, check=True)
+    # Pytest closes stdin by default, but some tests might need it to be open.
+    # In the future, should the need arise, we can feed custom data to stdin.
+    subprocess.run(kcov_cmd, shell=True, check=True, input=b'')
 
     # Read the coverage reported by kcov.
     coverage_file = os.path.join(kcov_output_dir, 'index.js')


integration_tests/test_coverage.py
```diff
@@ -79,7 +79,9 @@ def _get_current_coverage(coverage_config):
         exclude_pattern
     )
 
-    subprocess.run(kcov_cmd, shell=True, check=True)
+    # Pytest closes stdin by default, but some tests might need itto be open.
```
andreeaflorescu, line 82: typo: itto

integration_tests/test_coverage.py
```diff
@@ -79,7 +79,9 @@ def _get_current_coverage(coverage_config):
...
+    # In the future, should the need arise, we can feed custom data to stdin.
+    subprocess.run(kcov_cmd, shell=True, check=True, input=b'')
```
andreeaflorescu, line 84: I looked through the documentation, but cannot really understand what is the impact of having input=empty byte. It seems like this means that stdin will be a `PIPE`? Really not sure what is going on there. Could you expand on what happens (not in the code, but just in a comment here :)))

aghecenco: `stdin` won't be a `PIPE`, it will be a stream composed of the bytes passed here through `input=`. Without specifying `stdin=<something>` in the `subprocess.run` call, `stdin` gets closed by default. Some tests (looking at you, `vmm-reference` with stdin serial) expect `stdin` to be open, even if there's nothing coming through. Since the behavior before was to close it, passing an empty byte stream doesn't break existing behavior while circumventing the problem.

andreeaflorescu (approved):

lauralt (approved):

PR #49, opened 2020-10-14
buildkite: Skip cargo audit check temporarily

sameo:
Until our container moves to cargo audit v0.12.0.
See https://github.com/RustSec/advisory-db/issues/414#issuecomment-702479926

Signed-off-by: Samuel Ortiz <sameo@linux.intel.com>

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -207,6 +207,7 @@ steps:
   - label: "cargo-audit"
     commands:
       - cargo audit -q
+    skip: "Skipping until we move cargo audit to v0.12.0. See https://github.com/RustSec/advisory-db/issues/414#issuecomment-702479926"
     retry:
       automatic: false
     agents:


aghecenco (dismissed):

.buildkite/pipeline.yml
```diff
@@ -206,7 +206,7 @@ steps:
 
   - label: "cargo-audit"
     commands:
-      - cargo audit -q
+      - echo "Skipping cargo audit for now, see https://github.com/RustSec/advisory-db/issues/414#issuecomment-702479926" # cargo audit -q
```
andreeaflorescu, line 209: Can you use the `skip` field instead? You can check an example here: https://github.com/rust-vmm/rust-vmm-ci/blob/c8cf2b70dc21030172f77d5ac3c579c72eacc76e/.buildkite/pipeline.yml#L182

sameo: Nice. I changed it now.

andreeaflorescu (approved):

aghecenco (approved):

Issue #41, opened 2020-09-24
Make test commit work with repositories specified with `git`


andreeaflorescu:
[`test_commit_format`](https://github.com/rust-vmm/rust-vmm-ci/blob/master/integration_tests/test_commit_format.py) forever hangs when the repository is specified with `git` instead of `https`.
The reason it hangs is that doing `git fetch` in the docker container awaits user input to fetch from the unknown host (in this case github). A quick fix here is to run `ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts` before calling `git fetch`. This way at least the test will just fail, and not hang.

The second problem is that we need to forward the ssh keys on the host for git to work (and configure the ssh agent). No idea how to do that in a safe/sane way.

PR #56, opened 2020-12-08
don't run commit test for repos specified with git

lauralt:
Commit format test hangs if the repository is specified with git, we should skip the test for these repos until we find a safe way to run the test for these too (see #41 for more details about why is not safe).
Test is still running for repos with `https`: https://buildkite.com/rust-vmm/vm-superio-ci/builds/33


andreeaflorescu (approved):

andreeaflorescu:
This solution is soooo nice :+1: :100:

aghecenco (approved):

alexandruag (approved):

Issue #64, opened 2021-06-02
Expose vhost vdpa device to container


aghecenco:
To unblock https://github.com/rust-vmm/vhost/pull/33, several conditions must be met:
1. CI kernel version should be >= 5.7 (currently 5.4)
2. vhost vdpa kernel modules should be loaded (see https://github.com/rust-vmm/vhost/pull/33#issuecomment-843225243)
3. the `/dev/vhost-vdpa-0` device should be exposed to the container (see https://github.com/rust-vmm/vhost/pull/33#issuecomment-806000344)

1 and 2 can be fixed by updating the AMIs for the CI runners. This issue tracks step 3, for which we need to update the Buildkite pipeline so that the steps that involve running the unit tests can access the device. The device needs to be passed in the Buildkite yaml as `devices=[ "/dev/vhost-vdpa-0" ]` (see [documentation](https://github.com/buildkite-plugins/docker-buildkite-plugin#devices-optional-array)).

PR #65, opened 2021-06-02
vhost vdpa: expose device to container

aghecenco:
This PR exposes the `/dev/vhost-vdpa-0` device to the test container, to unblock testing vhost-vdpa support properly (see https://github.com/rust-vmm/vhost/pull/33). 

Do not merge until the CI images are upgraded so that the `/dev/vhost-vdpa-0` device exists on the host.

#64

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -80,6 +80,7 @@ steps:
           always-pull: true
           volumes:
             - "/tmp:/tmp"
+          devices: [ "/dev/vhost-vdpa-0" ]
 
   - label: "unittests-gnu-arm"
     commands:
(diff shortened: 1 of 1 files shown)


andreeaflorescu:
I think we might need to take a step back and think about how we want to support this in the future. It looks a bit odd to need to make the rust-vmm-ci device specific. What I have in mind is maybe thinking about the possibility for the rust-vmm-ci pipeline to be configurable, allowing the individual CIs to set additional parameters somehow.

Depending on how hard that might actually be, we can take the shortcut in this PR temporarily.

andreeaflorescu:
This is no longer needed because we're adding the changes in vhost directly. We still need to figure out a nicer way to pass custom configurations to rust-vmm-ci. I'll open a separate issue for that.

PR #67, opened 2021-06-21
check commit authors for each commit

keiichiw:
Although we check commit authors in test_commit_format.py to skip tests
for dependabot's commits, the current code only checks the author of the
head commit.
This PR make it check each commit's author so that we can test branches
where both a dependabot's commit and other commits exist.

Signed-off-by: Keiichi Watanabe <keiichiw@chromium.org>

### DIFF
--- a/integration_tests/test_commit_format.py
+++ b/integration_tests/test_commit_format.py
@@ -32,7 +32,7 @@ def test_commit_format():
 
     for sha in shas.split():
         # Do not enforce the commit rules when the committer is dependabot.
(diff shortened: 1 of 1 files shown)


keiichiw:
This is an improvement of code added by https://github.com/rust-vmm/rust-vmm-ci/pull/62. This change is useful in cases where we need to submit a dependabot's commit with some other commits at the same time like https://github.com/rust-vmm/vhost/pull/48.

lauralt (approved):

andreeaflorescu (approved):

PR #70, opened 2021-07-05
autogenerate pipeline from json

andreeaflorescu:
Instead of having a static pipeline.yml, autogenerate it from a JSON
configuration. The json approach is taken because we also want to be
able to run tests locally, and thus the json can be parsed once for
generating the pipeline, and one for generating a local test run (as
shown in test_run.py). For all the local tests to pass we also need to
adjust test_commit_format.py to not require Buildkite environment
variables.

This is still WIP (need to write docs).

### DIFF
--- a/.buildkite/generate_pipeline.py
+++ b/.buildkite/generate_pipeline.py
@@ -1,0 +1,115 @@
+#!/usr/bin/env python3
+
+"""
+This script is printing the Buildkite pipeline.yml to stdout.
+This can also be used as a library to print the steps from a different pipeline
+specified as a parameter to the `generate_test_pipeline`.
+
+The pipeline is generated based on the test configuration in
+`test_description.json`. The JSON contains list of tests to be run by all
+rust-vmm components.
+
+Some components need to override the default configuration such that they can
+access devices while running the tests (for example access to `/dev/kvm`),
+access to a temporary volume, and others. As such, this script supports
+overriding the following configurations through environment variables:
+- `X86_LINUX_AGENT_TAGS`: overrides the tags by which the x86_64 linux agent is
+  selected. The default tags are `platform: x86_64.metal, os: linux`.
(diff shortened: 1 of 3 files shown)


andreeaflorescu:
Tested with: https://github.com/andreeaflorescu/kvm-ioctls/pull/5

andreeaflorescu:
To run the tests locally, you just need to call the `test_run.py` script.

To run the Buildkite CI using the dynamic pipeline, you need to update the Buildkite configuration and run:
```./rust-vmm-ci/.buildkite/generate_pipeline.py | buildkite-agent pipeline upload```

The generate_pipeline.py allows overriding some value & extending some others through environment variables. For example, we can extend the docker plugin specification as follows:
```DOCKER_PLUGIN_CONFIG="devices: [ "/dev/kvm" ]" ./rust-vmm-ci/.buildkite/generate_pipeline.py | buildkite-agent pipeline upload```

All these are going to be properly documented once we agree on the high level design.

File: .buildkite/generate_pipeline.py [hunk 1]
```diff
@@ -0,0 +1,115 @@
...
+    for test in test_list["test_list"]:
+        for platform in test["platform"]:
+            label = "{}-{}".format(test['test_name'], platform)
+            command = test['command'].replace("{target_platform}", platform)
+            conditional = test.get("conditional", "")
+            buildkite_conditional = f"    if: {conditional}\n" if conditional else ""
+            print(
+                f"  - label: \"{label}\"\n"
+                "    commands:\n"
+                f"     - {command}\n"
```
catalindumitru, line 98: There's only 1 space here. Shouldn't there be 2?
```suggestion
                "    commands:\n"
                f"      - {command}\n"
```

andreeaflorescu:
Changes implemented in #76. Closing this.

PR #72, opened 2021-07-09
Update test_commit_format.py to work locally

catalindumitru:
Set default values for the base repository and base branch when the test is run locally and the Buildkite variables are not available.

### DIFF
--- a/integration_tests/test_commit_format.py
+++ b/integration_tests/test_commit_format.py
@@ -1,6 +1,14 @@
 # Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
-"""Test the commit message format."""
+"""
+Test the commit message format.
+
+This test works properly on the local machine only when the environment
+variables BASE_BRANCH and REMOTE are set. Otherwise the default values
+are "master" for the name of the base branch and "origin" for the
+remote name of the upstream repository, and this test may not work as
+expected.
+"""
 
 import os
 import subprocess
@@ -9,8 +17,14 @@ from utils import get_cmd_output
 
 COMMIT_TITLE_MAX_LEN = 50
 COMMIT_BODY_LINE_MAX_LEN = 72
-BASE_BRANCH = os.environ['BUILDKITE_PULL_REQUEST_BASE_BRANCH']
-BASE_REPO = os.environ['BUILDKITE_REPO']
+BASE_BRANCH = \
+    os.environ.get('BUILDKITE_PULL_REQUEST_BASE_BRANCH') or \
+    os.environ.get('BASE_BRANCH') or \
+    "master"
+REMOTE = \
+    os.environ.get('BUILDKITE_REPO') or \
+    os.environ.get('REMOTE') or \
+    "origin"
 
 
 def test_commit_format():
@@ -22,8 +36,14 @@ def test_commit_format():
     and if commits are signed.
     """
     # Fetch the upstream repository.
-    fetch_base_cmd = "git fetch {} {}".format(BASE_REPO, BASE_BRANCH)
-    subprocess.run(fetch_base_cmd, shell=True, check=True)
+    fetch_base_cmd = "git fetch {} {}".format(REMOTE, BASE_BRANCH)
+    try:
+        subprocess.run(fetch_base_cmd, shell=True, check=True)
+    except subprocess.CalledProcessError:
+        raise NameError(
+            "The name of the base branch or remote is invalid. "
+            "See test documentation for more details."
+        ) from None
     # Get hashes of PR's commits in their abbreviated form for
     # a prettier printing.
     shas_cmd = "git log --no-merges --pretty=%h --no-decorate " \


integration_tests/test_commit_format.py
```diff
@@ -3,14 +3,13 @@
...
 
-from utils import get_cmd_output
+from utils import get_cmd_result
 
 COMMIT_TITLE_MAX_LEN = 50
 COMMIT_BODY_LINE_MAX_LEN = 72
-BASE_BRANCH = os.environ['BUILDKITE_PULL_REQUEST_BASE_BRANCH']
-BASE_REPO = os.environ['BUILDKITE_REPO']
+BASE_BRANCH = os.environ.get('BUILDKITE_PULL_REQUEST_BASE_BRANCH', 'master')
+BASE_REPO = os.environ.get('BUILDKITE_REPO', 'origin')
```
lauralt, line 12: I would document a bit the default values, including when this test can be run locally, something like: `The commit format test works properly on the local machine only when the targeted branch of the pull request is "master" and the remote name of the upstream repository is "origin". If the targeted branch is for example a feature branch, then this test won't work as expected. In the rust-vmm CI of each repository these limitations no longer exist.`

integration_tests/test_commit_format.py
```diff
@@ -22,22 +21,24 @@ def test_commit_format():
     and if commits are signed.
     """
     # Fetch the upstream repository.
-    fetch_base_cmd = "git fetch {} {}".format(BASE_REPO, BASE_BRANCH)
-    subprocess.run(fetch_base_cmd, shell=True, check=True)
+    fetch_base_cmd = "git fetch -q {} {}".format(BASE_REPO, BASE_BRANCH)
+    fetch_stderr = get_cmd_result(fetch_base_cmd, result_type='stderr')
+    assert fetch_stderr == "", \
+        "Remote name should be origin. Base branch name should be master."
```
lauralt, line 27: The `should` word is not really accurate. The base branch can be a feature branch, in that case the user shouldn't change its name to master or something like that. The "origin, master" constraint is a test limitation, therefore the error message should reflect that.

integration_tests/utils.py
```diff
@@ -13,9 +13,11 @@ def get_repo_root_path():
     return os.path.dirname(rust_vmm_ci_path)
 
 
-def get_cmd_output(cmd):
-    """Returns stdout content of `cmd` command."""
-    cmd_out = subprocess.run(cmd, shell=True, check=True,
-                             stdout=subprocess.PIPE)
-    stdout = cmd_out.stdout.decode('utf-8')
-    return stdout
+def get_cmd_result(cmd, result_type='stdout'):
```
lauralt, line 16: I find returning different things based on a function argument a bit confusing. Moreover if you use as the `result_type` value something that is different from `stdout` or `stderr`, then the stdout content will be returned, which is not necessarily right. It would be clearer to return both values, i.e. `return stdout, stderr` at the end.
In this particular case you can also pass [ `check=True` to `subprocess.run()`](https://docs.python.org/3/library/subprocess.html#subprocess.run) and the test will immediately fail (with a decent message that I guess can also be improved) when the command fails.

integration_tests/test_commit_format.py
catalindumitru: Done.

lauralt (dismissed):

integration_tests/test_commit_format.py
```diff
@@ -1,16 +1,25 @@
 # Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
-"""Test the commit message format."""
+"""
+Test the commit message format.
+
+This test works properly on the local machine only when the base branch
```
andreeaflorescu, line 6: Working with origin is a bit of an anti-pattern. For example, locally I have the upstream remote named `upstream`, and `origin` points to my fork such that it is not possible to push to upstream by mistake (this in theory should not be possible because it should be configured in GitHub that you cannot push to master without reviews).

I would like us to be able to override these values with something that is not BUILDKITE_, but we can do that as a subsequent change.

integration_tests/test_commit_format.py
andreeaflorescu: Can we also rename this to `BASE_REMOTE` or something else closer to what we actually use it for?

integration_tests/test_commit_format.py
```diff
@@ -23,7 +32,15 @@ def test_commit_format():
     """
     # Fetch the upstream repository.
     fetch_base_cmd = "git fetch {} {}".format(BASE_REPO, BASE_BRANCH)
-    subprocess.run(fetch_base_cmd, shell=True, check=True)
+    try:
+        subprocess.run(fetch_base_cmd, shell=True, check=True)
+    except subprocess.CalledProcessError:
+        print(
```
andreeaflorescu, line 38: Should we use re-raise a more appropriate Exception instead? There are some details here about the best practices: https://stackoverflow.com/questions/2052390/manually-raising-throwing-an-exception-in-python

integration_tests/test_commit_format.py
catalindumitru: Done.

integration_tests/test_commit_format.py
catalindumitru: Done.

integration_tests/test_commit_format.py
catalindumitru: Done.

andreeaflorescu (approved):

lauralt (approved):

PR #73, opened 2021-07-20
Generate pipeline from json configuration

catalindumitru:
This is a similar approach to the one [here](https://github.com/rust-vmm/rust-vmm-ci/pull/70). The main difference is that this one uses pyyaml, which takes care of indenting the yaml output. This also includes a format checker of the json configuration.


.buildkite/pipeline.yml
```diff
@@ -1,231 +0,0 @@
-# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
-# SPDX-License-Identifier: Apache-2.0
```
andreeaflorescu, line 2: We will need to keep this file so that we are not forced to update all rust-vmm components at once. Once we roll out the new rust-vmm-ci to all components, we can add a PR to delete this file.

.buildkite/generate_pipeline.py
```diff
@@ -0,0 +1,96 @@
...
+DOCKER_PLUGIN_CONFIG = os.getenv('DOCKER_PLUGIN_CONFIG')
+
+PLATFORMS = ['x86_64.metal', 'arm.metal']
+OSS = ['linux']
+
+ROOT_PATH = pathlib.Path(__file__).parent.resolve()
+
+
+def check_step(step):
+    assert step.get('label'),\
```
lauralt, line 21: We can define a list of required keys to avoid the duplicate code. Also we can use a dictionary for the keys that accept only some particular values.

.buildkite/generate_pipeline.py
```diff
@@ -0,0 +1,96 @@
...
+        "Label missing."
+    assert step.get('commands'),\
+        "Commands missing."
+    agents = step.get('agents')
+    assert agents,\
+        "Agents missing."
+    assert agents.get('platform') in PLATFORMS,\
```
lauralt, line 28: I remembered that the platform was not actually required, e.g. the commit format test can be run on any of the available platforms. So we could check the value only if 'platform' was provided by the user. @andreeaflorescu what do you think?

.buildkite/generate_pipeline.py
```diff
@@ -0,0 +1,96 @@
...
+        "Plugins missing."
+
+
+def override_default_config(step):
+    agents = step['agents']
+    platform = agents['platform']
+
+    if X86_AGENT_TAGS and platform == 'x86_64.metal':
+        agents.clear()
+        tags = X86_AGENT_TAGS.split("; ")
```
lauralt, line 44: What happens if the values of the env variables don't match the expected format?
Can we match these to a regex before the actual split so that we can log a nicer message in case of an error?

Also looks like we could avoid the duplicate code here as well by defining some method.

.buildkite/generate_pipeline.py
```diff
@@ -0,0 +1,96 @@
...
+        ", ".join(PLATFORMS)
+    assert agents.get('os') in OSS,\
+        "Invalid or missing os. The supported oss are " + \
+        ", ".join(OSS)
+    assert step.get('plugins'),\
+        "Plugins missing."
+
+
+def override_default_config(step):
+    agents = step['agents']
```
lauralt, line 39: For usages like this, it is nice to leave a comment why we expect this to return a value.

PR #75, opened 2021-07-28
Setup for rust-vmm-ci-ci

catalindumitru:
In order to be able to run the CI on rust-vmm-ci we needed to
make it a rust crate. Also the path to the coverage config
file needed to be updated in test_coverage.py so that the path
does not depend on the crate; a breadth-first search was used
to guarantee that the config file belongs to the crate that is
being tested.

In the other crates, the tests reside in rust-vmm-ci/intergation_tests.
When we test the CI itself, the tests reside in integration_tests.
Therefore we needed to update the pytest commands to find the tests 
independently of the crate they are run from.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -172,7 +172,7 @@ steps:
 
   - label: "coverage-x86"
     commands:
-      - pytest rust-vmm-ci/integration_tests/test_coverage.py
+      - find . -type f -name "test_coverage.py" | xargs pytest
     retry:
       automatic: false
     agents:
@@ -205,7 +205,7 @@ steps:
 
   - label: "commit-format"
     commands:
-      - pytest rust-vmm-ci/integration_tests/test_commit_format.py
+      - find . -type f -name "test_commit_format.py" | xargs pytest
     if: build.env("BUILDKITE_REPO") !~ /^git@/
     retry:
       automatic: false
--- a/coverage_config_x86_64.json
+++ b/coverage_config_x86_64.json
@@ -1,0 +1,5 @@
+{
+    "coverage_score": 33.3, 
+    "exclude_path": "", 
+    "crate_features": ""
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,0 +1,5 @@
+[package]
+name = "rust-vmm-ci"
+version = "0.1.0"
+
+[dependencies]
--- a/integration_tests/test_coverage.py
+++ b/integration_tests/test_coverage.py
@@ -7,11 +7,29 @@ import pytest
 
 from utils import get_repo_root_path
 
+
+def get_coverage_config_path():
+    machine = platform.machine()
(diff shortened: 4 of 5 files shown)


.buildkite/pipeline.yml
```diff
@@ -172,7 +172,7 @@ steps:
 
   - label: "coverage-x86"
     commands:
-      - pytest rust-vmm-ci/integration_tests/test_coverage.py
+      - find . -type f -name "test_coverage.py" | xargs pytest
```
andreeaflorescu, line 175: The changes to the pipeline should be in a separate commit and we should explain in the commit message why are they needed.

coverage_config_aarch64.json
```diff
@@ -0,0 +1,5 @@
+{
```
andreeaflorescu, line 1: The aarch64 coverage test is not yet running. We can remove this file and add it only when we need it.

integration_tests/utils.py
```diff
@@ -13,6 +13,21 @@ def get_repo_root_path():
     return os.path.dirname(rust_vmm_ci_path)
 
 
+def get_coverage_config_path(platform):
```
andreeaflorescu, line 16: Since this is not a utility we expect to use in other tests beside the coverage one, we can add this function in the test_coverage.py file.

integration_tests/utils.py
```diff
@@ -13,6 +13,21 @@ def get_repo_root_path():
...
+    target_file = f"coverage_config_{platform}.json"
+    dirs = [os.getcwd()]
```
andreeaflorescu, line 18: Let's add a comment here why we are not using os.walkdir so that in the future people don't try to wrongly optimize it.

coverage_config_aarch64.json
catalindumitru: Done.

integration_tests/utils.py
catalindumitru: Done.

.buildkite/pipeline.yml
catalindumitru: Done.

integration_tests/utils.py
catalindumitru: Done.

andreeaflorescu (approved):

coverage_config_x86_64.json
```diff
@@ -0,0 +1,5 @@
+{
+    "coverage_score": 33.3,
```
lauralt, line 2: Why 33.3 if we have no tests? I also just ran `find . -type f -name "test_commit_format.py" | xargs pytest` in container and the reported value seems to be 0.

lauralt: Hmm, looks like it depends on the Rust version; from 1.49.0 the reported coverage is 33.3 and before -> 0. Anyway it's not related to the changes in this PR, so disregard my comment.

lauralt (approved):

PR #76, opened 2021-08-09
Autogenerate pipeline from json config file

catalindumitru:
Instead of having a static pipeline.yml, autogenerate it from a JSON
configuration file. The json approach is taken because we also want to
be able to run tests locally, and thus the json can be parsed once for
generating the pipeline, and once for generating a local test run.

### DIFF
--- a/.buildkite/autogenerate_pipeline.py
+++ b/.buildkite/autogenerate_pipeline.py
@@ -1,0 +1,255 @@
+#!/usr/bin/env python3
+
+"""
+This script is printing the Buildkite pipeline.yml to stdout.
+This can also be used as a library to print the steps from a different pipeline
+specified as a parameter to the `generate_test_pipeline`.
+
+The pipeline is generated based on the test configuration in
+`test_description.json`. The JSON file contains a list of tests to be run by
+all rust-vmm components.
+
+Some components need to override the default configuration such that they can
+access devices while running the tests (for example access to `/dev/kvm`),
+access to a temporary volume, and others. As such, this script supports
+overriding the following configurations through environment variables:
+- `X86_LINUX_AGENT_TAGS`: overrides the tags by which the x86_64 linux agent is
+  selected.
+- `AARCH64_LINUX_AGENT_TAGS`: overrides the tags by which the aarch64 linux
+  agent is selected.
+- `DOCKER_PLUGIN_CONFIG`: specifies additional configuration for the docker
+  plugin. For available configuration, please check the
+  https://github.com/buildkite-plugins/docker-buildkite-plugin.
+
+NOTE: The environment variables are specified as dictionaries, where the first
+key is `tests` and its value is a list of test names where the configuration
+should be applied; the second key is `cfg` and its value is a dictionary with
+the actual configuration.
+
+Examples of a valid configuration:
+```shell
+DOCKER_PLUGIN_CONFIG='{
+    "tests": ["coverage"],
+    "cfg": {
+        "devices": [ "/dev/vhost-vdpa-0" ],
+        "privileged": true
+    }
+}'
+```
+"""
+
+import yaml
+import json
+import os
+import sys
+import pathlib
+import copy
+
+# This represents the version of the rust-vmm-container used
+# for running the tests.
+CONTAINER_VERSION = "v12"
+# This represents the version of the Buildkite Docker plugin.
+DOCKER_PLUGIN_VERSION = "v3.8.0"
+
+X86_AGENT_TAGS = os.getenv('X86_LINUX_AGENT_TAGS')
+AARCH64_AGENT_TAGS = os.getenv('AARCH64_LINUX_AGENT_TAGS')
+DOCKER_PLUGIN_CONFIG = os.getenv('DOCKER_PLUGIN_CONFIG')
+
+PARENT_DIR = pathlib.Path(__file__).parent.resolve()
+
+
+class BuildkiteStep:
+    """
+    This builds a Buildkite step according to a json configuration and the
+    environment variables `X86_LINUX_AGENT_TAGS`, `AARCH64_LINUX_AGENT_TAGS`
+    and `DOCKER_PLUGIN_CONFIG`. The output is a dictionary.
+    """
+
+    def __init__(self):
+        """
+        Initialize a Buildkite step with default values.
+        """
+
+        # Default values.
+        # The order in which the attributes are initialized is the same as the
+        # order in which the keys will appear in the YAML file, because Python
+        # dictionaries are ordered. For readability reasons, this order should
+        # not be changed.
+        self.label = None
+        self.command = None
+        self.retry = {'automatic': False}
+        self.agents = {'os': 'linux'}
+        self.plugins = [
+            {
+                f"docker#{DOCKER_PLUGIN_VERSION}": {
+                    'image': f"rustvmm/dev:{CONTAINER_VERSION}",
+                    'always-pull': True
+                }
+            }
+        ]
+
+    def _set_platform(self, platform):
+        """ Set platform if given in the json input. """
+
+        if platform:
+            # We need to change `aarch64` to `arm` because of the way we are
+            # setting the tags on the host.
+            if platform == 'aarch64':
+                platform = 'arm'
+            self.agents['platform'] = f"{platform}.metal"
+
+    def _set_conditional(self, conditional):
+        """ Set conditional if given in the json input. """
+
+        if conditional:
+            setattr(self, 'if', conditional)
+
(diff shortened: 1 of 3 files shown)


.buildkite/test_description.json
```diff
@@ -0,0 +1,61 @@
...
+      "platform": ["x86_64"]
+    },
+    {
+      "test_name": "commit-format",
+      "command": "find . -type f -name \"test_commit_format.py\" | xargs pytest",
+      "conditional": "build.env(\"BUILDKITE_REPO\") !~ /^git@/",
+      "docker_plugin": {
+        "propagate-environment": true
+      },
+      "platform": ["x86_64"]
```
lauralt, line 53: Is this still needed?

.buildkite/autogenerate_pipeline.py
```diff
@@ -0,0 +1,230 @@
...
+
+            for platform in platforms:
+                step_input = copy.deepcopy(test)
+                step_input['platform'] = platform
+
+                step = BuildkiteStep()
+                step_output = step.build(step_input)
+                self.steps.append(step_output)
+
+        return self.__dict__
```
lauralt, line 214: Can we leave a comment explaining what __dict__ does?

.buildkite/test_description.json
catalindumitru: Removed the platform key for the tests that don't need it.

.buildkite/autogenerate_pipeline.py
catalindumitru: Done. Also replaced `self.__dict__` with `vars(self)`. This seems to be the more "Pythonic" way because it avoids calling magic methods such as `__dict__`.

.buildkite/autogenerate_pipeline.py
```diff
@@ -0,0 +1,240 @@
...
+"""
+
+import yaml
+import json
+import os
+import sys
+import pathlib
+import copy
+
+CONTAINER_VERSION = "v12"
```
andreeaflorescu, line 46: We should add a comment about this const to say that it represents the rust-vmm-container used for running the tests.

Same for the other const, we should say that it represents the Buildkite docker plugin version.

.buildkite/autogenerate_pipeline.py
```diff
@@ -0,0 +1,240 @@
...
+    environment variables `X86_LINUX_AGENT_TAGS`, `AARCH64_LINUX_AGENT_TAGS`
+    and `DOCKER_PLUGIN_CONFIG`. The output is a dictionary.
+    """
+
+    def __init__(self):
+        """
+        Initialize a Buildkite step with default values.
+        """
+
+        # Default values
```
andreeaflorescu, line 68: If I remember correctly, the order in which we define these matters. We should add a comment in the code here to make sure that people don't change the order by mistake.

.buildkite/autogenerate_pipeline.py
```diff
@@ -0,0 +1,240 @@
...
+                    'image': f"rustvmm/dev:{CONTAINER_VERSION}",
+                    'always-pull': True
+                }
+            }
+        ]
+
+    def _set_platform(self, platform):
+        """ Set platform if given in the json input. """
+
+        if platform:
```
andreeaflorescu, line 85: Can you add a comment here why you need to do this? (it's a problem for the way we are setting the tag on the host, it should've been aarch64; nevertheless we should explain it)

.buildkite/autogenerate_pipeline.py
```diff
@@ -0,0 +1,240 @@
...
+        """
+        Override the tags by which the linux agent is selected
+        using the `X86_LINUX_AGENT_TAGS` and `AARCH64_LINUX_AGENT_TAGS`
+        environment variables.
+        """
+
+        env_var = None
+        platform = self.agents.get('platform')
+
+        if platform:
```
andreeaflorescu, line 137: Can you add a comment about the logic here? It only becomes clear what you want to do when reading the build function. We should say that when no platform is specified, we don;t add the platform to the dictionary so that the test can run in any env.

andreeaflorescu (commented):
LGTM high level. We should add some information in the readme about how this works now, and why we ended up implementing it this way.

.buildkite/autogenerate_pipeline.py
catalindumitru: Done.

.buildkite/autogenerate_pipeline.py
catalindumitru: Done.

.buildkite/autogenerate_pipeline.py
catalindumitru: Done.

.buildkite/autogenerate_pipeline.py
catalindumitru: Done.

andreeaflorescu (dismissed):

README.md
```diff
@@ -49,15 +49,42 @@ The json must have the following fields:
...
+
+For example, we can extend the docker plugin specification as follows:
+```shell
+DOCKER_PLUGIN_CONFIG='{
+    "tests": ["coverage"],
+    "cfg": {
+        "devices": [ "/dev/vhost-vdpa-0" ],
+        "privileged": true
+    }
+}' ./rust-vmm-ci/.buildkite/autogenerate_pipeline.py | buildkite-agent pipeline upload
```
andreeaflorescu, line 84: nit: should this be on the following line?

README.md
```diff
@@ -49,15 +49,42 @@ The json must have the following fields:
...
+4. There is a script that autogenerates a dynamic Buildkite pipeline. To run 
+the CI using this dynamic pipeline, you need to add a step that is uploading
+the rust-vmm-ci pipeline:
 ```bash
-buildkite-agent pipeline upload rust-vmm-ci/.buildkite/pipeline.yml
+./rust-vmm-ci/.buildkite/autogenerate_pipeline.py | buildkite-agent pipeline upload
+```
+This allows overriding some values and extending others through environment 
+variables. 
+- `X86_LINUX_AGENT_TAGS`: overrides the tags by which the x86_64 linux agent is
```
andreeaflorescu, line 63: Let's also write about the defaults and say that for most use cases overriding these is not necessary.

We might also want to give a few examples when we do want to override them (platform needs a custom device that is not available on the existing test instances, need for a specialized hypervisor - for example, microsoft hyper-v).

README.md
```diff
@@ -49,15 +49,42 @@ The json must have the following fields:
 This file is required for the coverage integration so it needs to be added
 to the repository as well.
 
-1. Create a new pipeline definition in Buildkite. For this step ask one of the
-rust-vmm Buildkite [admins](CODEOWNERS) to create one for you. Add a pipeline
-step that is uploading the rust-vmm-ci pipeline:
+3. Create a new pipeline definition in Buildkite. For this step ask one of the
```
andreeaflorescu, line 52: Can you add a link to the process from the community repo? Here is the link: https://github.com/rust-vmm/community/blob/master/docs/maintainers/setup_new_repo.md#set-up-ci

README.md
catalindumitru: Yes, changed it.

README.md
catalindumitru: Done.

README.md
catalindumitru: Added the link.

andreeaflorescu (dismissed):

README.md
```diff
@@ -49,15 +49,52 @@ The json must have the following fields:
...
+rust-vmm Buildkite [admins](CODEOWNERS) to create one for you. The process is explained
+[here](https://github.com/rust-vmm/community/blob/master/docs/maintainers/setup_new_repo.md#set-up-ci).
 
+4. There is a script that autogenerates a dynamic Buildkite pipeline. To run 
+the CI using this dynamic pipeline, you need to add a step that is uploading
+the rust-vmm-ci pipeline:
 ```bash
-buildkite-agent pipeline upload rust-vmm-ci/.buildkite/pipeline.yml
+./rust-vmm-ci/.buildkite/autogenerate_pipeline.py | buildkite-agent pipeline upload
```
lauralt, line 60: I would update at lest the example from #Buildkite Pipeline section with how it will look after running autogenerate_pipeline.py? (fewer newlines etc.).

lauralt (commented):
Looks good! I'll approve once we manage to test this end to end.

README.md
catalindumitru: Updated the other examples.

lauralt (approved):

andreeaflorescu (approved):

PR #78, opened 2021-08-20
Add script to run tests locally

catalindumitru:
The script uses the test_description.json file to get the test names and the commands to be run.


andreeaflorescu (changes requested):
We need to add a readme entry about running the tests locally. We have something vague about running the coverage test, but it would be better to replace it with running all the tests.

In the readme we should also specify that the tests should run with the container, and point somehow to where the latest version of the container is defined, so that people know how to run it.


Running this locally on vm-fdt yields the following errors:

```bash
test_coverage (__main__.TestsContainer) ... Traceback (most recent call last):
...
    raise ValueError("option names %s already added" % conflict)
ValueError: option names {'--profile'} already added
```

```bash
test_commit-format (__main__.TestsContainer) ... Traceback (most recent call last):
..
    raise ValueError("option names %s already added" % conflict)
ValueError: option names {'--profile'} already added
```

Weirdly enough, even if there are these errors, the tests still show as passed.

andreeaflorescu (dismissed):

test_run.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python3
```
lauralt, line 1: Please add copyright.

catalindumitru: Added it.

PR #79, opened 2021-08-20
Allow skipping tests when generating the pipeline

catalindumitru:
Added a new environment variable for skipping tests if needed when generating the pipeline.

### DIFF
--- a/.buildkite/autogenerate_pipeline.py
+++ b/.buildkite/autogenerate_pipeline.py
@@ -11,8 +11,9 @@ all rust-vmm components.
 
 Some components need to override the default configuration such that they can
 access devices while running the tests (for example access to `/dev/kvm`),
-access to a temporary volume, and others. As such, this script supports
-overriding the following configurations through environment variables:
+access to a temporary volume, and others. Some components may also need to skip
+some of the tests. As such, this script supports overriding the following
+configurations through environment variables:
 - `X86_LINUX_AGENT_TAGS`: overrides the tags by which the x86_64 linux agent is
   selected.
 - `AARCH64_LINUX_AGENT_TAGS`: overrides the tags by which the aarch64 linux
@@ -20,14 +21,17 @@ overriding the following configurations through environment variables:
 - `DOCKER_PLUGIN_CONFIG`: specifies additional configuration for the docker
   plugin. For available configuration, please check the
   https://github.com/buildkite-plugins/docker-buildkite-plugin.
+- `TESTS_TO_SKIP`: specifies a list of tests to be skipped.
 
-NOTE: The environment variables are specified as dictionaries, where the first
-key is `tests` and its value is a list of test names where the configuration
-should be applied; the second key is `cfg` and its value is a dictionary with
-the actual configuration.
+NOTE: The variable `TESTS_TO_SKIP` is specified as a JSON list with the names
+of the tests to be skipped. The other variables are specified as dictionaries,
+where the first key is `tests` and its value is a list of test names where the
+configuration should be applied; the second key is `cfg` and its value is a
+dictionary with the actual configuration.
 
 Examples of a valid configuration:
 ```shell
+TESTS_TO_SKIP='["commit-format"]'
 DOCKER_PLUGIN_CONFIG='{
     "tests": ["coverage"],
     "cfg": {
@@ -54,6 +58,7 @@ DOCKER_PLUGIN_VERSION = "v3.8.0"
 X86_AGENT_TAGS = os.getenv('X86_LINUX_AGENT_TAGS')
 AARCH64_AGENT_TAGS = os.getenv('AARCH64_LINUX_AGENT_TAGS')
 DOCKER_PLUGIN_CONFIG = os.getenv('DOCKER_PLUGIN_CONFIG')
+TESTS_TO_SKIP = os.getenv('TESTS_TO_SKIP')
 
 PARENT_DIR = pathlib.Path(__file__).parent.resolve()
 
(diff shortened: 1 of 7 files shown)


integration_tests/test_coverage.py
```diff
@@ -125,12 +131,12 @@ def test_coverage(profile, no_cleanup):
...
             # In the CI Profile we expect the coverage to be manually updated.
-            assert False, "Coverage is increased from {} to {}. " \
-                          "Please update the coverage in " \
-                          "tests/coverage.".format(
-                previous_coverage,
-                current_coverage
-            )
+            assert False,\
+                "Coverage is increased from {} to {}. "\
+                "Please update the coverage in tests/coverage.".format(
```
andreeaflorescu, line 136: Since you are updating this, can you also specify the correct path instead of tests/coverage?

integration_tests/conftest.py
```diff
@@ -36,6 +45,16 @@ def no_cleanup(request):
     return request.config.getoption("--no-cleanup")
 
 
+@pytest.fixture
+def base_branch(request):
+    return request.config.getoption("--base_branch")
```
andreeaflorescu, line 50: Can we name this `--base-branch` instead to have the same naming as for the existing `no-cleanup` option?

.buildkite/autogenerate_pipeline.py
```diff
@@ -54,6 +54,7 @@
 X86_AGENT_TAGS = os.getenv('X86_LINUX_AGENT_TAGS')
 AARCH64_AGENT_TAGS = os.getenv('AARCH64_LINUX_AGENT_TAGS')
 DOCKER_PLUGIN_CONFIG = os.getenv('DOCKER_PLUGIN_CONFIG')
+TESTS_TO_SKIP = os.getenv('TESTS_TO_SKIP')
```
andreeaflorescu, line 57: I think we should add something about this in the readme as well.

integration_tests/conftest.py
```diff
@@ -21,8 +21,17 @@ def pytest_addoption(parser):
         "--no-cleanup",
         action="store_true",
         default=False,
-        help="Keep the coverage report in `kcov_output` directory. If this flag is not provided, "
-             "both coverage related directories are removed."
+        help="Keep the coverage report in `kcov_output` directory. If this "
+             "flag is not provided, both coverage related directories are "
+             "removed."
+    )
+    parser.addoption(
```
andreeaflorescu, line 28: I find it a bit strange to have this configuration here because this parameter is actually valid just for one test. Of course, same goes for `--no-cleanup`. Can we at least add a TODO to find a better way to provide parameters to tests?

catalindumitru: I don't like the current approach either, but couldn't find an alternative that keeps using pytest. Added a TODO to look for another way.

.buildkite/autogenerate_pipeline.py
catalindumitru: Done.

integration_tests/conftest.py
catalindumitru: Yes, changed it.

integration_tests/test_coverage.py
catalindumitru: Updated the path.

integration_tests/conftest.py
```diff
@@ -21,8 +24,17 @@ def pytest_addoption(parser):
...
+        "--base-branch",
+        default="master"
```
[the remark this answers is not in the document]
lauralt: Sounds good to me.

.buildkite/test_description.json
```diff
@@ -17,35 +23,48 @@
...
-      "platform": ["x86_64", "aarch64"]
+      "platform": [
+        "x86_64",
+        "aarch64"
+      ]
     },
     {
       "test_name": "coverage",
-      "command": "find . -type f -name \"test_coverage.py\" | xargs pytest",
+      "command": "pytest $(find . -type f -name \"test_coverage.py\") --remote {REMOTE_NAME} --base-branch {BASE_BRANCH}",
```
andreeaflorescu, line 57: This needs to be updated.

integration_tests/test_coverage.py
```diff
@@ -133,7 +133,8 @@ def test_coverage(profile, no_cleanup):
             # In the CI Profile we expect the coverage to be manually updated.
             assert False,\
                 "Coverage is increased from {} to {}. "\
-                "Please update the coverage in tests/coverage.".format(
+                "Please update the coverage in coverage_config_x86_64.json "\
+                "and coverage_config_aarch64.json.".format(
```
lauralt, line 137: Can we use `platform.machine()` instead of hardcoding the platform?

integration_tests/test_benchmark.py
```diff
@@ -1,16 +1,20 @@
 # Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
-"""Compare benchmark results before and after a pull request."""
+"""
+Compare benchmark results before and after a pull request.
+
+This test works properly on the local machine only when the options
+`--base-branch` and `--remote` are set. Otherwise the default values
+are "master" for the name of the base branch and "origin" for the
```
lauralt, line 8: "master" will have to be updated if we switch to the `main` default.

integration_tests/test_coverage.py
catalindumitru: Yes, that's better. Changed it.

integration_tests/test_benchmark.py
catalindumitru: I think we should stick to "master" until the update.

andreeaflorescu (approved):

lauralt (approved):

integration_tests/test_coverage.py
```diff
@@ -125,12 +131,13 @@ def test_coverage(profile, no_cleanup):
...
             # In the CI Profile we expect the coverage to be manually updated.
-            assert False, "Coverage is increased from {} to {}. " \
-                          "Please update the coverage in " \
-                          "tests/coverage.".format(
-                previous_coverage,
-                current_coverage
-            )
+            assert False,\
+                "Coverage is increased from {} to {}. "\
+                "Please update the coverage in coverage_config_{}.".format(
```
lauralt, line 136: This is missing the `.json` part actually.

PR #81, opened 2021-08-31
Add help option for scripts

catalindumitru:
Added help options `-h, --help` for the script that autogenerates
the pipeline and for the one that runs the tests locally. In the
former, replaced the module `optparse` (which is now [deprecated](https://docs.python.org/3/library/optparse.html))
with the module `argparse`.

### DIFF
--- a/.buildkite/autogenerate_pipeline.py
+++ b/.buildkite/autogenerate_pipeline.py
@@ -51,7 +51,8 @@ import sys
 import pathlib
 import copy
 
-from optparse import OptionParser
+from argparse import ArgumentParser, RawTextHelpFormatter
+from textwrap import dedent
 
 # This represents the version of the rust-vmm-container used
 # for running the tests.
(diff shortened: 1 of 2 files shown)


andreeaflorescu (dismissed):

test_run.py
```diff
@@ -31,6 +34,19 @@ def retrieve_test_list(
 
 
 if __name__ == '__main__':
+    help_text = dedent(
+        """
+        The tests "test_benchmark.py" and "test_commit_format.py" work properly
```
lauralt, line 39: This is starting too abruptly I think. I would begin with that the script does, i.e. lets you run all the tests locally in one shot or something like that.

.buildkite/autogenerate_pipeline.py
```diff
@@ -267,14 +268,30 @@ def generate_pipeline(config_file):
...
     # that have custom pipelines, and it helps with keeping the container
     # version the same across pipelines.
-    parser.add_option("-t", "--test-description", dest="test_description",
-                      help="JSON file containing the test description.",
-                      default=f"{PARENT_DIR}/test_description.json")
-    (options, args) = parser.parse_args()
-    generate_pipeline(options.test_description)
+    parser.add_argument('-t', '--test-description',
+                        metavar="JSON_FILE",
+                        help='JSON file containing the test description.',
```
lauralt, line 294: I would make the help message clearer, something like: the path to the json file containing the test description for the CI.

lauralt (approved):

andreeaflorescu (approved):

PR #85, opened 2021-09-06
Remove static pipeline

catalindumitru:
The use of the static file `pipeline.yml` is now deprecated. All pipelines should use the dynamic script `autogenerate_pipeline.py`.


andreeaflorescu (changes requested):
Let's wait to first merge the dependabot update so we don't get spammed with PRs daily again.

lauralt (approved):
I just noticed that [this section](https://github.com/rust-vmm/rust-vmm-ci/blob/main/README.md#custom-pipeline) is a bit outdated, but we can update it in a separate PR.

lauralt (approved):

andreeaflorescu (approved):

PR #87, opened 2021-09-15
Add timeout for Buildkite steps 

catalindumitru:
Added a default timeout value for each Buildkite step that can be overridden through the environment variable `TIMEOUTS` and the JSON configuration file using the key "timeout_in_minutes" in each step.

### DIFF
--- a/.buildkite/autogenerate_pipeline.py
+++ b/.buildkite/autogenerate_pipeline.py
@@ -24,12 +24,15 @@ configurations through environment variables:
   plugin. For available configuration, please check the
   https://github.com/buildkite-plugins/docker-buildkite-plugin.
 - `TESTS_TO_SKIP`: specifies a list of tests to be skipped.
+- `TIMEOUTS_MIN`: overrides the timeout value for specific tests.
 
 NOTE: The variable `TESTS_TO_SKIP` is specified as a JSON list with the names
-of the tests to be skipped. The other variables are specified as dictionaries,
-where the first key is `tests` and its value is a list of test names where the
-configuration should be applied; the second key is `cfg` and its value is a
-dictionary with the actual configuration.
+of the tests to be skipped. The variable `TIMEOUTS_MIN` is a dictionary where
+each key is the name of a test and each value is the number of minutes for the
+timeout. The other variables are specified as dictionaries, where the first key
+is `tests` and its value is a list of test names where the configuration should
+be applied; the second key is `cfg` and its value is a dictionary with the
+actual configuration.
 
 Examples of a valid configuration:
 ```shell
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -73,17 +73,20 @@ variables.
...
-specification as follows:
+of the tests to be skipped. The variable `TIMEOUTS` is a dictionary where each
+key is the name of a test and each value is the number of minutes for the
+timeout. The other variables are specified as dictionaries, where the first key
+is `tests` and its value is a list of test names where the configuration should
+be applied; the second key is `cfg` and its value is a dictionary with the
+actual configuration.
+
+For example, we can skip the test `commit-format`, have a timeout of 30 minutes
+for the test `style` and extend the docker plugin specification as follows:
```
lauralt, line 87: The example is not using the `style` test in TIMEOUTS.

.buildkite/autogenerate_pipeline.py
```diff
@@ -122,6 +129,12 @@ def _add_docker_config(self, cfg):
             for key, val in cfg.items():
                 target[key] = val
 
+    def _set_timeout(self, timeout):
```
lauralt, line 132: I would keep only the env variable setting.

.buildkite/autogenerate_pipeline.py
```diff
@@ -24,12 +24,15 @@
   plugin. For available configuration, please check the
   https://github.com/buildkite-plugins/docker-buildkite-plugin.
 - `TESTS_TO_SKIP`: specifies a list of tests to be skipped.
+- `TIMEOUTS`: overrides the timeout value for specific tests
```
lauralt, line 27: nit: full stop.

.buildkite/autogenerate_pipeline.py
catalindumitru: Removed the method.

README.md
catalindumitru: Changed it, thanks!

.buildkite/autogenerate_pipeline.py
```diff
@@ -41,6 +44,7 @@
         "privileged": true
     }
 }'
+TIMEOUTS='["commit-format": 10]'
```
andreeaflorescu, line 47: A good practice for variables that have measurement units is to write the unit in the name. This would remove the need to wonder what is it measured in, and makes the code easier to use. The timeout can be renamed to `TIMEOUTS_MIN`.

catalindumitru: Renamed it, thanks!

lauralt (dismissed):

lauralt (approved):

andreeaflorescu (approved):

PR #89, opened 2021-09-23
update container version to v13

andreeaflorescu:
In the new container version we're using ubuntu 20.04 as the base image
instead of ubuntu 18.04.


lauralt (approved):
Not sure if we should keep the examples in README up to date with the container version.

andreeaflorescu:
Good idea, I'll update it.

lauralt (approved):

lauralt (approved):

jiangliu (approved):

Issue #74, opened 2021-07-23
coverage test


andreeaflorescu:
Make sure that there are no spaces in the JSON field that defines the features, otherwise we can encounter failures when running kcov.

Example of failure: https://buildkite.com/rust-vmm/mshv-ci/builds/7#cd3a2709-ad84-484a-9fef-ab41da00b7c8

andreeaflorescu:
We should also check that it is a valid json if that is possible. Do you think you could also add a check for that @ahmedaabouzied?

PR #93, opened 2021-10-11
Assert no spaces in crate features JSON

ahmedaabouzied:
Spaces in the `crate_features` JSON value can cause errors when running `kcov`.
 
The `json.load()` method used to load the JSON file will raise an error if it's not valid JSON as mentioned on the python [json module docs](https://docs.python.org/3/library/json.html)
> If the data being deserialized is not a valid JSON document, a JSONDecodeError will be raised.

Fixes: #74

### DIFF
--- a/integration_tests/test_coverage.py
+++ b/integration_tests/test_coverage.py
@@ -51,6 +51,9 @@ def _read_test_config():
     assert "exclude_path" in coverage_config
     assert "crate_features" in coverage_config
 
+    assert ' ' not in coverage_config["crate_features"], \
+        "spaces are not allowed in crate_features value"
+
     return coverage_config
 
 


integration_tests/test_coverage.py
```diff
@@ -51,6 +51,9 @@ def _read_test_config():
     assert "exclude_path" in coverage_config
     assert "crate_features" in coverage_config
 
+    assert coverage_config["crate_features"].find(' ') == -1, \
```
andreeaflorescu, line 54: nit: I find the syntax `' ' not in coverage_config["crate_features"]` to be more descriptive than using `-1` for the check.

ahmedaabouzied: Good point. Updated that.

PR #102, opened 2022-02-11
make coverage test less sensitive to changes

andreeaflorescu:
Instead of accounting for any difference in coverage, only fail the test
if the difference between the current coverage and the new coverage is
higher than 0.5%.

### DIFF
--- a/integration_tests/test_coverage.py
+++ b/integration_tests/test_coverage.py
@@ -135,26 +135,27 @@ def test_coverage(profile, no_cleanup, test_scope):
     coverage_config = _read_test_config()
     current_coverage = _get_current_coverage(coverage_config, no_cleanup, test_scope)
     previous_coverage = coverage_config["coverage_score"]
-    if previous_coverage < current_coverage:
-        if profile == pytest.profile_ci:
-            # In the CI Profile we expect the coverage to be manually updated.
-            assert False, \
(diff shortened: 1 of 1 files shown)


andreeaflorescu:
0.5% was chosen based on data from the latest 19 PRs in kvm-ioctls. More details here: https://github.com/rust-vmm/community/pull/127#discussion_r804554453

integration_tests/test_coverage.py
```diff
@@ -135,26 +135,27 @@ def test_coverage(profile, no_cleanup, test_scope):
...
-        else:
-            # This should never happen because pytest should only accept
-            # the valid test profiles specified with `choices` in
-            # `pytest_addoption`.
-            assert False, "Invalid test profile."
-    elif previous_coverage > current_coverage:
-        diff = float(previous_coverage - current_coverage)
-        assert False, "Coverage drops by {:.2f}%. Please add unit tests for " \
-                      "the uncovered lines.".format(diff)
+    diff = float(previous_coverage - current_coverage)
```
lauralt, line 138: Is this returning the absolute value of the result? Should we use `abs` here?

andreeaflorescu: We should use `abs` good catch!

lauralt (approved):

sameo (approved):

PR #105, opened 2022-08-26
update to container version 16

andreeaflorescu:
### Summary of the PR

This version adds some tooling for fuzzing and updated to Rust 1.63.0.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/.buildkite/autogenerate_pipeline.py
+++ b/.buildkite/autogenerate_pipeline.py
@@ -60,7 +60,7 @@ from textwrap import dedent
 
 # This represents the version of the rust-vmm-container used
 # for running the tests.
-CONTAINER_VERSION = "v15"
+CONTAINER_VERSION = "v16"
 # This represents the version of the Buildkite Docker plugin.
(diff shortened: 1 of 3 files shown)


lauralt (approved):
nit: we could also update the docs so that they point to the latest version published. It might actually be a good idea to replace the hardcoded values with v${LATEST} everywhere as we already do and explain [here](https://github.com/rust-vmm/rust-vmm-ci#running-the-tests-locally).

lauralt (approved):

andreeaflorescu:
> nit: we could also update the docs so that they point to the latest version published. It might actually be a good idea to replace the hardcoded values with v${LATEST} everywhere as we already do and explain [here](https://github.com/rust-vmm/rust-vmm-ci#running-the-tests-locally).

I've updated the README as well and added a note that the version specified there might not be the latest one. We might forget to update it from time to time.

lauralt (approved):

alexandruag (approved):

PR #113, opened 2022-11-29
audit: ignore unmaintained atty advisory

aghecenco:
### Summary of the PR

`cargo audit` is currently failing due to [this advisory](https://rustsec.org/advisories/RUSTSEC-2021-0145) in the (apparently) unmaintained `atty` crate. Ignoring this specific advisory in order to unblock CI.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/.buildkite/test_description.json
+++ b/.buildkite/test_description.json
@@ -71,7 +71,7 @@
     },
     {
       "test_name": "cargo-audit",
-      "command": "cargo audit -q --deny warnings"
+      "command": "cargo audit -q --deny warnings --ignore RUSTSEC-2021-0145"
     }
   ]
 }


.buildkite/test_description.json
```diff
@@ -71,7 +71,7 @@
     },
     {
       "test_name": "cargo-audit",
-      "command": "cargo audit -q --deny warnings"
+      "command": "cargo audit -q --deny warnings --ignore RUSTSEC-2021-0145"
```
andreeaflorescu, line 74: I think it might make more sense to have this in the audit.toml file instead and only apply it to the crates where it makes sense. If there are multiple crates for which this happens, we can also have it in the `.github` repository as this is what will end up being linked in all repositories when a file with the same name does not exist.

I prefer the approach with audit.toml because it makes it possible to define it per repository when repositories have different needs. I am not expecting the list to be ever growing, but maybe it might make sense in the future to ignore the vulnerabilities for some crates, but not for others. One example might be when a dependency is used for testing for a crate (thus safe-ish to ignore) vs a normal dependency (in most cases not safe to ignore).

What do you think?

andreeaflorescu: I am not 100% sure that linking audit.toml via .github will work though, we need to invest a bit of time in this. Here is the repo I am talking about: https://github.com/rust-vmm/.github.

andreeaflorescu:
We are now ignoring this per repository. Closing this PR.

PR #116, opened 2022-11-30
Reduce technical debt & increase timeout

andreeaflorescu:
### Summary of the PR

We were using a rather outdated version. Moving to the latest available one.
There are some other things that need fixing and I am looking at addressing in this PR, hence for now this is WIP.

### Requirements

requirements:

- [X] All commits in this PR are signed (with `git commit -s`), and the commit
  description line.
  test.

### DIFF
--- a/.buildkite/autogenerate_pipeline.py
+++ b/.buildkite/autogenerate_pipeline.py
@@ -62,7 +62,7 @@ from textwrap import dedent
 # for running the tests.
 CONTAINER_VERSION = "v16"
 # This represents the version of the Buildkite Docker plugin.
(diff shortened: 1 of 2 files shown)


andreeaflorescu:
Another thing that I want to do before merging this is to publish a new container version to update the base image from Ubuntu 18.04 to ubuntu 22.

andreeaflorescu:
> Another thing that I want to do before merging this is to publish a new container version to update the base image from Ubuntu 18.04 to ubuntu 22.

This will take longer than anticipated, so I would propose just merging this once we get the hosts in order as well.

jiangliu (approved):

lauralt (approved):

PR #120, opened 2023-04-03
update to container v21

vireshk:
That updates to a newer version of Rust, 1.68.2, and adds back libclang-dev package which is required by vhost-device/gpio crates.

### Summary of the PR

### Requirements

requirements:

- [ ] All commits in this PR are signed (with `git commit -s`), and the commit
  description line.
  test.

### DIFF
--- a/.buildkite/autogenerate_pipeline.py
+++ b/.buildkite/autogenerate_pipeline.py
@@ -60,7 +60,7 @@ from textwrap import dedent
 
 # This represents the version of the rust-vmm-container used
 # for running the tests.
-CONTAINER_VERSION = "v19"
+CONTAINER_VERSION = "v21"
 # This represents the version of the Buildkite Docker plugin.
 DOCKER_PLUGIN_VERSION = "v5.3.0"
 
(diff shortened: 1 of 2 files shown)


.buildkite/autogenerate_pipeline.py
```diff
@@ -60,7 +60,7 @@
 
 # This represents the version of the rust-vmm-container used
 # for running the tests.
-CONTAINER_VERSION = "v19"
+CONTAINER_VERSION = "v21"
```
andreeaflorescu, line 63: How come we skipped one version?

andreeaflorescu (commented):
Because of https://github.com/rust-vmm/rust-vmm-container/issues/79 we will not be able to update the rust-vmm-ci because it will cause intermittent failures on the cargo audit test. For now, can you update the audit test to only run on x86_64 until we identify what the problem is? To do that, you will need to update the `cargo audit` configuration in .buildkite/test_description.json

.buildkite/autogenerate_pipeline.py
vireshk: V20 was createad a weak back when you merged this I guess:

https://github.com/rust-vmm/rust-vmm-container/pull/77

andreeaflorescu (dismissed):

lauralt (approved):

andreeaflorescu (approved):

PR #130, opened 2023-07-07
test_commit_format: no longer check body length

Ablu:
There are quite a few, genuine useful cases where one wants to overflow the recommended 75 (some use 72) character rule.

Breaking long lines or manually breaking long compiler or other tool-generated outputs is just confusing and error-prone.

This does not mean that we should stop wrapping manually written text at a sane length, but enforcing the rule with a script seems to yield questionable results while creating quite a bit of friction.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/integration_tests/test_commit_format.py
+++ b/integration_tests/test_commit_format.py
@@ -28,9 +28,8 @@ def test_commit_format():
     """
     Checks commit message format for the current PR's commits.
 
-    Checks if commit messages follow the 60/75 commit rule (a maximum
-    60 characters for the title and 75 characters for description
-    lines) and if commits are signed.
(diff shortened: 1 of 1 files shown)


epilys (approved):
The argument for URLs and error messages is a pragmatic one, since there's no good way to do line breaking on those.

A good compromise would be printing a warning, when using this as a pre-commit git hook.

roypat (approved):
I'm definitely not married to this test. I feel like machines that can only display 80 characters in a line went out of style long before I was even born. For just the URL things, theres ways to work around it by adjusting the gitlint rules, but the error message length thing is something I also run into constantly, and breaking them apart is often awkward. +1 from me.

JonathanWoollett-Light (approved):

PR #131, opened 2023-07-17
test_run.py: some QoL tweaks to the test_run script

stsquad:
This adds some simple QoL improvements for those trying to run individual tests from the command line.

### Requirements

requirements:

  description line.
- [n/a] All added/changed functionality has a corresponding unit/integration
  test.
- [n/a] All added/changed public-facing functionality has entries in the "Upcoming 
- [n/a] Any newly added `unsafe` code is properly documented.


test_run.py
```diff
@@ -66,4 +76,5 @@ def retrieve_test_list(config_file=f"{PARENT_DIR}/.buildkite/test_description.js
             setattr(TestsContainer, f"test_{name}", test_func)
 
     if not args.list_tests:
-        unittest.main(verbosity=2)
+        tests_suite = unittest.TestLoader().loadTestsFromTestCase(TestsContainer)
+        unittest.TextTestRunner(verbosity=2).run(tests_suite)
```
Ablu, line 80: Why is this needed? Why is the `if` + `continue` not enough above?

stsquad: Fun story - I'm not sure. I was running into weird errors where it was erroring out and I couldn't figure out why. I had a chat (http://ix.io/4API) with gpt4 which suggested the change which worked. It's reasoning might be off but it seemed to imply that the unittests had already been parsed meaning the updated attributes didn't take effect. However I don't understand the guts of the unit test framework enough. The loading of the tests_suite seemed sane though.

Ablu: I do not fully understand why it worked before and no longer does... The GPT explanation does not make sense to me, since the class starts off empty in both cases.

Maybe this is a bit easier?: https://paste.centos.org/view/8c66c3ef. Not sure if the `__name__` assignment is a lot better though...

PR #132, opened 2023-08-29
chore: Update container version to 25

kalyazin:
### Summary of the PR

Update container version to 25

### Requirements

requirements:

- [ ] All commits in this PR are signed (with `git commit -s`), and the commit
  description line.
  test.

### DIFF
--- a/.buildkite/autogenerate_pipeline.py
+++ b/.buildkite/autogenerate_pipeline.py
@@ -60,7 +60,7 @@ from textwrap import dedent
 
 # This represents the version of the rust-vmm-container used
 # for running the tests.
-CONTAINER_VERSION = "v24"
+CONTAINER_VERSION = "v26"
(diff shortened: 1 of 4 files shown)


roypat (approved):

JonathanWoollett-Light (dismissed):

roypat:
The coverage failure is because with rust 1.72.0, `cargo kcov` broke. It relies on parsing rustc's stdout for discovering where test executable are located, and the format for that changed (cargo-kcov expects the log message to start with "Running rustc (...)", but since 1.72.0 cargo prints out the fully qualified path to rustc, resulting in cargo-kcov not finding any binaries). Since `cargo-kcov` is unmaintained since 2019, I'll go ahead and replace it with grcov.

roypat (approved):

JonathanWoollett-Light (approved):

PR #137, opened 2023-10-19
Fix `exclude path` parsing and update documentation.

stefano-garzarella:
### Summary of the PR

We don't want the shell to interpret the contents of exclude_path,
but we want it all used as a single string for `--ignore-filename-regex`,
so let's put it between double quotes.

Let's also update the `exclude_path` description to prevent multiple files
from being separated by comma.

### Requirements

requirements:

  description line.
  test.


roypat (approved):
Thanks for fixing this!

stefano-garzarella:
I forgot to update the examples! I will do in v2.

integration_tests/test_coverage.py
```diff
@@ -80,7 +80,7 @@ def _get_current_coverage(coverage_config, no_cleanup, test_scope):
 
     additional_exclude_path = coverage_config["exclude_path"]
     if additional_exclude_path:
-        llvm_cov_command += f" --ignore-filename-regex {additional_exclude_path}"
+        llvm_cov_command += f" --ignore-filename-regex \"{additional_exclude_path}\""
```
Ablu, line 83: And all of that just because the code uses `shell=True` instead of just arranging the parameters in an array :/

Ablu (approved):

coverage_config_aarch64.json.sample
```diff
@@ -1,5 +1,5 @@
 {
   "coverage_score": 90,
-  "exclude_path": "my_autogenerated_file.rs,second_file.rs",
+  "exclude_path": "my_autogenerated_file.rs|second_file.rs",
```
roypat, line 3: I think technically the fullstops here should be escaped

roypat (approved):

coverage_config_aarch64.json.sample
stefano-garzarella: Do you mean for json parsing?

stefano-garzarella: If I use `virtio-bindings|mock\.rs` I need to escape `\` so it should be `virtio-bindings|mock\\.rs` , but I didn't see any issue using just `.`

roypat: Because `.` in regex matches anything, so this would match files named `second_file[any character]rs`. Although maybe not the biggest problem, give that such files are highly unlikely to actually contain code

stefano-garzarella: @roypat oh, I was forgetting about the regular expression, yes you are right! I'll update ;-)

stefano-garzarella:
v3:
- escaped fullstops and added a path to a folder in the examples

roypat (approved):

JonathanWoollett-Light (approved):

PR #139, opened 2023-10-20
Generate Cargo.lock before call `cargo audit`

stefano-garzarella:
### Summary of the PR

Starting from v0.18.0, cargo-audit hangs indefinitely if Cargo.lock does not exist. We discovered this while upgrading the container from v26 to v28 [1], which among other things updated cargo-audit.

For the binary crates this should not be a problem, since they have Cargo.lock committed, but for many libraries this may not be true.

If Cargo.lock is not there, we are generating one with the latest available versions, which may not be very significant. For this and other reasons it's now suggested that libraries also have a Cargo.lock [2] committed (thanks Manos for pointing this out).

Note: `cargo generate-lockfile` updates Cargo.lock if it's already there, but we don't want it, that's why we have the guard.

[1] https://github.com/rust-vmm/rust-vmm-ci/pull/138
[2] https://doc.rust-lang.org/nightly/cargo/faq.html#why-have-cargolock-in-version-control

Suggested-by: @epilys 
Suggested-by: @roypat 

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/.buildkite/test_description.json
+++ b/.buildkite/test_description.json
@@ -71,7 +71,7 @@
     },
     {
       "test_name": "cargo-audit",
-      "command": "cargo audit -q --deny warnings",
+      "command": "[ -e Cargo.lock ] || cargo generate-lockfile; cargo audit -q --deny warnings",
       "platform": [
         "x86_64"
       ]


stefano-garzarella:
As we discussed in #138 , another option would be to skip cargo audit if there is no committed Cargo.lock or force to have one for libraries as well (but this last point maybe better in the future when we are sure everyone has it).

epilys (approved):

roypat (approved):

PR #141, opened 2023-10-23
test_coverage: fix formatting of --ignore-filename-regex

stefano-garzarella:
### Summary of the PR

`black` suggests putting the string between single quotes so there is no need to escape double quotes.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/integration_tests/test_coverage.py
+++ b/integration_tests/test_coverage.py
@@ -82,7 +82,7 @@ def _get_current_coverage(coverage_config, no_cleanup, test_scope):
 
     additional_exclude_path = coverage_config["exclude_path"]
     if additional_exclude_path:
-        llvm_cov_command += f" --ignore-filename-regex \"{additional_exclude_path}\""
+        llvm_cov_command += f' --ignore-filename-regex "{additional_exclude_path}"'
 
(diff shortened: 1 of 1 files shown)


roypat (approved):

stefano-garzarella:
@roypat should we run `black` on PRs?

roypat:
> @roypat should we run `black` on PRs?

I think it might make sense for this repository, but I'm not really a big fan of how it wraps lines. e.g. sometimes when you have a very long line that ends with a list or dictionary access, it will give you something like 

```py
very_long_thing[
    0
]
```

which... why?

stefano-garzarella:
> > @roypat should we run `black` on PRs?
> 
> I think it might make sense for this repository, but I'm not really a big fan of how it wraps lines. e.g. sometimes when you have a very long line that ends with a list or dictionary access, it will give you something like
> 
> ```python
> very_long_thing[
>     0
> ]
> ```
> 
> which... why?

yeah, I see!

I mentioned this because this is the second time after merging a PR, I get a notification that `black` fails, and as a python profane I didn't launch it locally before sending the previous PR :-(

andreeaflorescu (approved):

PR #159, opened 2024-08-02
Introduce `riscv64` CI

TimePrinciple:
### Summary of the PR

Introduce logic necessary for generating YAML needed by BuildKite, which are designed to work with image introduced in
rust-vmm/rust-vmm-container#106.

### Requirements

requirements:

  test.

### DIFF
--- a/.buildkite/autogenerate_pipeline.py
+++ b/.buildkite/autogenerate_pipeline.py
@@ -60,7 +60,12 @@ from textwrap import dedent
 
 # This represents the version of the rust-vmm-container used
 # for running the tests.
-CONTAINER_VERSION = "v38"
+CONTAINER_VERSION = "v44"
+# The suffix suggests that the dev image with `v{N}-riscv` tag is not to be
+# confused with real `riscv64` image (it's actually a `x86_64` image with
+# `qemu-system-riscv64` installed), since AWS yet has `riscv64` machines
+# available.
+CONTAINER_VERSION_RISCV = CONTAINER_VERSION + "-riscv"
 # This represents the version of the Buildkite Docker plugin.
 DOCKER_PLUGIN_VERSION = "v5.3.0"
 
@@ -78,8 +83,7 @@ TIMEOUTS_MIN = os.getenv("TIMEOUTS_MIN")
 # More details here: https://github.com/rust-vmm/community/issues/137
 DEFAULT_AGENT_TAG_HYPERVISOR = os.getenv("DEFAULT_AGENT_TAG_HYPERVISOR", "kvm")
 
-PARENT_DIR = pathlib.Path(__file__).parent.resolve()
-
+BUILDKITE_PATH = pathlib.Path(__file__).parent.resolve()
 
 class BuildkiteStep:
     """
@@ -237,6 +241,17 @@ class BuildkiteStep:
         if "{target_platform}" in command:
             assert platform, "Command requires platform, but platform is missing."
             command = command.replace("{target_platform}", platform)
+        # Modify command and tag name for `riscv64` CI
+        if platform == "riscv64":
+            # Wrap command with '' to avoid escaping early by `ENTRYPOINT`
+            command = json.dumps(command)
+            # Overwrite image tag for riscv64 platform CI
+            self.step_config["plugins"][0][f"docker#{DOCKER_PLUGIN_VERSION}"]["image"] = f"rustvmm/dev:{CONTAINER_VERSION_RISCV}"
+            # Since we are using qemu-system inside a x86_64 container, we
+            # should set `platform` field to x86_64 and unset the hypervisor to
+            # be passed
+            platform = "x86_64"
+            hypervisor = ""
         self.step_config["command"] = command
 
         # Optional keys.
@@ -287,7 +302,7 @@ class BuildkiteConfig:
     def __init__(self):
         self.bk_config = None
 
-    def build(self, input):
+    def build(self, input, platform_allowlist):
         """Build the final Buildkite configuration fron the json input."""
 
         self.bk_config = {"steps": []}
@@ -309,6 +324,11 @@ class BuildkiteConfig:
                 platforms = [None]
 
             for platform in platforms:
+                # Filter test enabled in platform_allowlist
+                if platform not in platform_allowlist:
+                    # Skip disabled platform
+                    continue
+
                 step_input = copy.deepcopy(test)
                 step_input["platform"] = platform
                 if not step_input.get("hypervisor"):
@@ -322,7 +342,18 @@ class BuildkiteConfig:
         return self.bk_config
 
 
-def generate_pipeline(config_file):
+def determine_allowlist(config_file):
+    """Determine the what platforms should be enabled for this crate"""
+
+    try:
+        with open(config_file, 'r') as file:
+            platforms = [line.strip() for line in file.readlines()]
+        return platforms
+    except Exception as e:
+        # Fall back to default platform if anything goes wrong
+        return ["x86_64", "aarch64"]
+
+def generate_pipeline(config_file, platform_allowlist):
     """Generate the pipeline yaml file from a json configuration file."""
 
     with open(config_file) as json_file:
@@ -330,7 +361,7 @@ def generate_pipeline(config_file):
         json_file.close()
 
     config = BuildkiteConfig()
-    output = config.build(json_cfg)
+    output = config.build(json_cfg, platform_allowlist)
     yaml.dump(output, sys.stdout, sort_keys=False)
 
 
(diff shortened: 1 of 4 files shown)


.buildkite/autogenerate_pipeline.py
```diff
@@ -363,5 +392,13 @@ def generate_pipeline(config_file):
         help="The path to the JSON file containing the test" " description for the CI.",
         default=f"{PARENT_DIR}/test_description.json",
     )
+    parser.add_argument(
+        "-p",
+        "--platform-allowlist",
+        metavar="PLATFORM_DOT_FILE",
+        help="The path to the dot file containing the allowed platforms for the CI.",
+        default=f"{REPO_ROOT_DIR}/.platform",
```
stefano-garzarella, line 400: What about using something similar of what we do for `coverage_config_{machine}.json` ?

See https://github.com/rust-vmm/rust-vmm-ci/blob/cec97d9e6c578949d40226b8b85fa4ad3142e31f/integration_tests/test_coverage.py#L16

.buildkite/autogenerate_pipeline.py
```diff
@@ -363,5 +392,13 @@ def generate_pipeline(config_file):
         help="The path to the JSON file containing the test" " description for the CI.",
         default=f"{PARENT_DIR}/test_description.json",
     )
+    parser.add_argument(
+        "-p",
+        "--platform-allowlist",
+        metavar="PLATFORM_DOT_FILE",
+        help="The path to the dot file containing the allowed platforms for the CI.",
```
stefano-garzarella, line 399: Maybe we should also document what happen if the file is not specified.

.buildkite/autogenerate_pipeline.py
```diff
@@ -61,6 +61,7 @@
 # This represents the version of the rust-vmm-container used
 # for running the tests.
 CONTAINER_VERSION = "v35"
+CONTAINER_VERSION_RISCV = CONTAINER_VERSION + "-riscv"
```
stefano-garzarella, line 64: Why we need `-riscv` in the tag?

TimePrinciple: In `docker-publish.yml` of rust-vmm/rust-vmm-ci#159, since the `riscv64` image is actually a `x86_64` image with `qemu-system-riscv64` installed, I think it is not as standard enough as the other two platforms (they are tested on actual hardware) to have a name like `dev:riscv64`, that would be kind of misleading, people would tend to think that's a `riscv64` image. So I suggest we use name like `v35-riscv` for now, and standardize it afterwords

stefano-garzarella: LGTM, maybe a comment there will be helpful

TimePrinciple: Noted, I will provide comments to them in next update

.buildkite/autogenerate_pipeline.py
```diff
@@ -361,7 +393,18 @@ def generate_pipeline(config_file):
...
         help="The path to the JSON file containing the test" " description for the CI.",
-        default=f"{PARENT_DIR}/test_description.json",
+        default=f"{BUILDKITE_PATH}/test_description.json",
+    )
+    parser.add_argument(
+        "-p",
+        "--platform-allowlist",
+        metavar="PLATFORM_DOT_FILE",
+        help=(
+            "The path to the dot file containing the allowed platforms for the CI. "
```
roypat, line 403: Let's document the format of this file somewhere (e.g. one architecture per line, and which architectures are supported in the first place). We'll also need to update the relevant parts of the README, which currently still states that CI will always run on x86_64 and ARM64 - here we should also mention the configurability now exposed via the .platforms file + its format.

TimePrinciple: Current logic would obey lines read from `.platform` if present, should I add some constraint to enable `x86_64` and `arm64` on any circumstances 🤔

TimePrinciple: > Current logic would obey lines read from `.platform` if present, should I add some constraint to enable `x86_64` and `arm64` on any circumstances 🤔

I think we should observe platforms provided by `.platform`, documentations are ready now

roypat (commented):
Thanks :) Just some grammar nits.

.buildkite/autogenerate_pipeline.py
```diff
@@ -78,8 +83,8 @@
 # More details here: https://github.com/rust-vmm/community/issues/137
 DEFAULT_AGENT_TAG_HYPERVISOR = os.getenv("DEFAULT_AGENT_TAG_HYPERVISOR", "kvm")
 
-PARENT_DIR = pathlib.Path(__file__).parent.resolve()
-
+BUILDKITE_PATH = os.path.dirname(os.path.realpath(__file__))
```
stefano-garzarella, line 86: I can't put a comment on line 55, but `pathlib` is not used anymore after this change.

BTW why are we switching from pathlib to os.path? (I'm not a python expert at all)
IMHO we should do that change in a separate commit, and explain why.

.buildkite/autogenerate_pipeline.py
```diff
@@ -237,6 +242,17 @@ def build(self, input):
...
+        if platform == "riscv64":
+            # Wrap command with '' to avoid escaping early by `ENTRYPOINT`
+            command = json.dumps(command)
+            # Overwrite image tag for riscv64
+            self.step_config["plugins"][0][f"docker#{DOCKER_PLUGIN_VERSION}"]["image"] = f"rustvmm/dev:{CONTAINER_VERSION_RISCV}"
+            # Set platform to x86_64 for riscv64
+            platform = "x86_64"
+            # Since we are using qemu-system inside the container, we don't need
+            # hypervisor to be passed here
+            hypervisor = ""
```
stefano-garzarella, line 255: I'd avoid superfluous comments that describe exactly what the code does. Comments should explain why or clarify obscure things, for example here we should explain that we are doing this because the container will be an x86_64 that will run qemu, etc.

.buildkite/autogenerate_pipeline.py
TimePrinciple: The `os.path` way is derived from `test_coverage.py`, I'm switching it back to `pathlib`

.buildkite/autogenerate_pipeline.py
TimePrinciple: Noted, I tried to improve the comment a bit, please have a look when convenient ：）

stefano-garzarella: much better now, thanks!

stefano-garzarella:
The PR LGTM, please rebase it.

TimePrinciple:
> The PR LGTM, please rebase it.

Thanks for reviewing, rebased now :)

stefano-garzarella (approved):

roypat (commented):
Let's wait for the rust-vmm-container CI to build the new containers, and then let's bump `CONTAINER_VERSION` here to get a final sanity check on everything :) But then we're good to go I think!

TimePrinciple:
> Let's wait for the rust-vmm-container CI to build the new containers, and then let's bump `CONTAINER_VERSION` here to get a final sanity check on everything :) But then we're good to do I think!

Shall I do it in this PR or raise another one 🧐

TimePrinciple:
> > > Let's wait for the rust-vmm-container CI to build the new containers, and then let's bump `CONTAINER_VERSION` here to get a final sanity check on everything :) But then we're good to do I think!
> > 
> > 
> > Shall I do it in this PR or raise another one 🧐
> 
> In this PR please :)

Container version updated to v39 :)

roypat:
> > > > Let's wait for the rust-vmm-container CI to build the new containers, and then let's bump `CONTAINER_VERSION` here to get a final sanity check on everything :) But then we're good to do I think!
> > > 
> > > 
> > > Shall I do it in this PR or raise another one 🧐
> > 
> > 
> > In this PR please :)
> 
> Container version updated to v39 :)

I think it'll need to be v40 (... which is my fault, I forgot to do a PR to update it to v39 when my rust toolchain update container got merged. Sorry!)

TimePrinciple:
> > > > > Let's wait for the rust-vmm-container CI to build the new containers, and then let's bump `CONTAINER_VERSION` here to get a final sanity check on everything :) But then we're good to do I think!
> > > > 
> > > > 
> > > > Shall I do it in this PR or raise another one 🧐
> > > 
> > > 
> > > In this PR please :)
> > 
> > 
> > Container version updated to v39 :)
> 
> I think it'll need to be v40 (... which is my fault, I forgot to do a PR to update it to v39 when my rust toolchain update container got merged. Sorry!)

Never mind, I've force pushed to v40 :)

roypat:
Mh, I just realized that it didn't even try to spawn the riscv64 CI jobs for _this_ PR. Can we add the .platform file to this repository, so that the CI sanity checks we have also always run on riscv64? I think for this we'll need to make the .platform file discovery code slightly more clever, see for instance how the coverage test does it: https://github.com/rust-vmm/rust-vmm-ci/blob/main/integration_tests/test_coverage.py#L16

TimePrinciple:
> Mh, I just realized that it didn't even try to spawn the riscv64 CI jobs for _this_ PR. Can we add the .platform file to this repository, so that the CI sanity checks we have also always run on riscv64? I think for this we'll need to make the .platform file discovery code slightly more clever, see for instance how the coverage test does it: https://github.com/rust-vmm/rust-vmm-ci/blob/main/integration_tests/test_coverage.py#L16

I gather that we need a mechanism to detect where this script is being executed, as a normal repo or submodule, and invoke CI accordingly. I will work it out.

roypat (approved):
:tada:

stefano-garzarella (approved):

stefano-garzarella:
I just opened https://github.com/rust-vmm/rust-vmm-ci/pull/162 to fix formatting issue highlighted by `black` on main: https://github.com/rust-vmm/rust-vmm-ci/actions/runs/10662874591/job/29550878188

PR #165, opened 2024-09-26
run unittests in privileged docker container

roypat:
The code coverage tests already run in a privileged container, so this does not affect our security posture. Additionally, it will unblock vsock unit tests in the vhost-user-vsock crate, see also https://github.com/rust-vmm/vhost-device/pull/706 and https://github.com/rust-vmm/vhost-device/pull/728

### Summary of the PR

### Requirements

requirements:

  test.

### DIFF
--- a/.buildkite/test_description.json
+++ b/.buildkite/test_description.json
@@ -28,7 +28,10 @@
         "x86_64",
         "aarch64",
         "riscv64"
-      ]
+      ],
+      "docker_plugin": {
+        "privileged": true
+      }
     },
     {
       "test_name": "unittests-musl",
@@ -36,7 +39,10 @@
       "platform": [
         "x86_64",
         "aarch64"
-      ]
+      ],
+      "docker_plugin": {
+        "privileged": true
+      }
     },
     {
       "test_name": "clippy",


TimePrinciple:
I was about to ask if we could do the same for `rust-vmm-container`, I'll need to mount directories inside container to create images since some tests would fail in `9p rootfs` as of `riscv64`.

I will bring up this discussion recently :)

stefano-garzarella (approved):

.buildkite/test_description.json
```diff
@@ -28,15 +28,21 @@
         "x86_64",
         "aarch64",
         "riscv64"
-      ]
+      ],
+      "docker_plugin": {
```
lauralt, line 32: Is it difficult to enable this only in the `vhost` related repositories where this is needed? I remember there was some mechanism to update such configuration for a particular repository, but I might be wrong.

TimePrinciple: I noticed in `vhost` repo there is a separate step which uses `rust-vmm-ci-test.json` to upload tests to be executed in that repo, as shown [here](https://buildkite.com/rust-vmm/vhost-ci/builds/805#01924d98-2e4e-42b0-a45b-63a420a9e4e2). But this is not used in `vhost-device` currently.

roypat: I think that configuration is only about adding more tests? e.g. we could add an _additional_ integration test that runs inside of a privileged container, and it executes the subset of unittests that require privileges. That does seem a bit convoluted though, wdyt @stefano-garzarella?

The advantage of enabling it globally is that it brings the coverage and unittest integration tests in-line (since coverage is already ran in a privileged container). Although maybe with the switch to llvm-cov a while ago, this is not needed anymore?

stefano-garzarella: In `vhost` repo we had to redefine `rust-vmm-ci-test.json` because we can't use `--all-features` since the `xen` feature is not compatible with some others features. In both `vhost` and `vhost-device` we have a custom pipeline to do more tests.

I agree with @roypat, currently here we run coverage tests privileged, but not unit tests, so I think we should decide whether to avoid privileges for all or not. Since we already do this for coverage, maybe the easiest and most painless thing is to do it for unit tests as well.

But if we decide to run everything without privileges, then it's perfectly fine for me to go and redefine the pipelines for those repos where we need them.

lauralt: Ok, makes sense 👍

lauralt (approved):

stefano-garzarella:
Just to confirm that now https://github.com/rust-vmm/vhost-device/pull/728 is working great with this PR merged!

PR #167, opened 2024-10-03
ci: Enalbe tests with no `platform` specified

TimePrinciple:
### Summary of the PR

After `.platform` mechanism was introduced in https://github.com/rust-vmm/rust-vmm-ci/pull/159, it silently filters
out tests with no `platform` specified. Enable them to execute by adding
`platform is not None` predicate preceeds allowlist check.

### Requirements

requirements:

  test.

### DIFF
--- a/.buildkite/autogenerate_pipeline.py
+++ b/.buildkite/autogenerate_pipeline.py
@@ -328,7 +328,7 @@ class BuildkiteConfig:
 
             for platform in platforms:
                 # Filter test enabled in platform_allowlist
-                if platform not in platform_allowlist:
+                if platform is not None and platform not in platform_allowlist:
                     # Skip disabled platform
                     continue
 


TimePrinciple:
Very sorry about breaking our CI tests 😭

roypat (approved):
Ah, well, its just the `commit-format` test. No big deal, thanks for noticing and fixing!

TimePrinciple:
> Ah, well, its just the `commit-format` test. No big deal, thanks for noticing and fixing!

And `style` test as well :(

roypat:
> > Ah, well, its just the `commit-format` test. No big deal, thanks for noticing and fixing!
> 
> And `style` test as well :(

Ah true. Still, out of all the tests it could've hit, these were definitely the least important ones, so don't beat yourself up over it! It's just as much our fault for not noticing it during review

.buildkite/autogenerate_pipeline.py
```diff
@@ -321,10 +321,10 @@ def build(self, input, platform_allowlist):
                 if test_name in tests_to_skip:
                     continue
 
-            # The platform is optional. When it is not specified, we don't add
-            # it to the step so that we can run the test in any environment.
+            # The platform is optional. When it is not specified, it defaults
+            # to `x86_64` to get checked against `platform_allowlist`.
             if not platforms:
-                platforms = [None]
+                platforms = ["x86_64"]
```
lauralt, line 327: Can't we make it run on any platform that is available instead (basically what the old comment states)? So that CI execution is potentially finishing faster.

.buildkite/autogenerate_pipeline.py
```diff
@@ -321,10 +321,10 @@ def build(self, input, platform_allowlist):
                 if test_name in tests_to_skip:
```
lauralt, line 321: nit: typo in commit title: `enalbe`

.buildkite/autogenerate_pipeline.py
TimePrinciple: Noted, I'll try to work it out :)

TimePrinciple: Updated, PTAL when convenient :)

.buildkite/autogenerate_pipeline.py
TimePrinciple: Fixed :)

lauralt (approved):

roypat (approved):

PR #171, opened 2024-10-23
fix: report line coverage instead of function coverage

### LINKED ISSUE #170: Is using "function coverage" metric instead of "line coverage" intended?
While working with @dorindabassey on a [gpu in vhost-device](https://github.com/rust-vmm/vhost-device/pull/668), she has noticed the CI seem to be using the "function coverage" metric of llvm-cov. Then I also noticed the comment in the python code of the CI, that seems to suggest "line coverage" is intended to be used.

https://github.com/rust-vmm/rust-vmm-ci/blob/209c04eb5ebed01b1dc3fbcfcb3ba2d563695773/integration_tests/test_coverage.py#L101-L106
The comment here doesn't seem to be correct - at least on my machine the order of the columns in the CLI output of `cargo llvm-cov test --summary-only` is: `Region coverage`, `Function coverage`, `Line coverage`.

Confusingly the order of the columns is different when using the html output (`cargo llvm-cov test --open`): `Function coverage`, `Line coverage`, `Region coverage`.

Btw there is also `cargo llvm-cov test --json --summary-only` which could prevent such confusion.

This means that it seems like the projects using this CI, use the function coverage metric instead of line coverage.
Is this intended or it is it a bug in the parsing of the output? I would think the original intention was to use line coverage.

roypat:
It seems that the raw output of llvm-cov changed at some point, and we ended upcomparing functoin coverage instead of line coverage. Fix this by instead using the new json output, which will hopefully prevent such goofs in the future.

Fixes #170
Signed-off-by: Patrick Roy <roypat@amazon.co.uk>

### Requirements

requirements:

  test.

### DIFF
--- a/integration_tests/test_coverage.py
+++ b/integration_tests/test_coverage.py
@@ -77,7 +77,7 @@ def _get_current_coverage(coverage_config, no_cleanup, test_scope):
     shutil.rmtree(cov_build_dir, ignore_errors=True)
 
     llvm_cov_command = (
-        f"CARGO_TARGET_DIR={cov_build_dir} cargo llvm-cov test --summary-only"
+        f"CARGO_TARGET_DIR={cov_build_dir} cargo llvm-cov test --json --summary-only"
     )
 
     additional_exclude_path = coverage_config["exclude_path"]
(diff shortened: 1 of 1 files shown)


roypat:
Maybe we could also talk about getting rid about these coverage files altogether and instead using codecov.io at some point? We did that at Firecracker a year ago or so

stefano-garzarella:
> Maybe we could also talk about getting rid about these coverage files altogether and instead using codecov.io at some point? We did that at Firecracker a year ago or so

Yeah, can you open an issue, so if someone has free cycle can look at it.

stefano-garzarella (approved):

ShadowCurse (approved):

PR #173, opened 2024-12-09
feat: run unit tests with --release flag

ShadowCurse:
### Summary of the PR

Rust does optimizations which can cause havoc in `--release` mode, so add another set of unit tests
with all optimizations enabled.

See: https://github.com/rust-vmm/kvm/pull/298

### Requirements

requirements:

  test.

### DIFF
--- a/.buildkite/test_description.json
+++ b/.buildkite/test_description.json
@@ -45,6 +45,29 @@
       }
     },
     {
+      "test_name": "unittests-gnu-release",
+      "command": "cargo test --release --all-features --workspace",
+      "platform": [
+        "x86_64",
+        "aarch64",
+        "riscv64"
+      ],
+      "docker_plugin": {
+        "privileged": true
+      }
+    },
+    {
+      "test_name": "unittests-musl-release",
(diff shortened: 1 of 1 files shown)


roypat (approved):

stefano-garzarella:
The changes LGTM, but I have a couple of questions:
1. should we run unit tests for both debug and release? (not sure if for the same reason, running tests only with optimizations, can hide some issues)
2. should we do the same for coverage? (in integration_tests/test_coverage.py we call `cargo llvm-cov test` without `--release`)

ShadowCurse:
Added unit tests with debug build back.
Regarding coverage, I would assume this is not impacted by the optimization level, so no reason to add it there.

stefano-garzarella:
> Added unit tests with debug build back.

Can you also update commit and PR title/description?

>Regarding coverage, I would assume this is not impacted by the optimization level, so no reason to add it there.

I see, definitely not to be fixed in this PR, but your changes made me wonder if it makes sense to calculate coverage by considering, for example, code that is not used in release (maybe we don't even have it, but it occurred to me). I'll open an issue to discusse that

stefano-garzarella (approved):

PR #177, opened 2025-02-17
rustvmm_gen: Introduce `rustvmm_gen`

RuoqingHe:
### Summary of the PR

- rustvmm_gen: Introduce lib_kernel_source.py
- rustvmm_gen: Introduce lib_syscall.py
- rustvmm_gen: Introduce rustvmm_gen

`rustvmm_gen.py` integrates kernel version checking, source extraction, compilation and header installation. Header files could be prepared with command:

```command
./rustvmm_gen/rustvmm_gen.py --arch <arm64/x86_64/riscv> --version <X.Y/X.Y.Z> prepare
```

Source code required by `seccompiler` could be generated with command:

```command
./rustvmm_gen/rustvmm_gen.py --arch <arm64/x86_64/riscv> --version <X.Y/X.Y.Z> generate_syscall
```

### Requirements

requirements:

  test.

### DIFF
--- a/scripts/rustvmm_gen.py
+++ b/scripts/rustvmm_gen.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python3
+#
+# Copyright 2025 © Institute of Software, CAS. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import argparse
+import os
+from pathlib import Path
+from lib.kernel_source import prepare_source
+from lib.seccompiler import generate_seccompiler, SECCOMPILER_SYSCALL_DIR
+from lib.kvm_bindings import generate_kvm_bindings, KVM_BINDINGS_DIR
+
+
+def main():
+    parser = argparse.ArgumentParser(prog="rustvmm_gen")
+    subparsers = parser.add_subparsers(dest="command", required=True)
+    parser.add_argument("--arch", help="Target architecture (x86_64, arm64, riscv)")
+    parser.add_argument("--version", required=True, help="Kernel version (e.g. 6.12.8)")
+    parser.add_argument(
+        "--install_path",
+        default=None,
+        help="Header installation directory path",
+    )
+    parser.add_argument("--keep", help="Keep temporary build files")
+
+    # Prepare subcommand
+    prepare_parser = subparsers.add_parser("prepare", help="Prepare kernel headers")
+    prepare_parser.set_defaults(func=prepare_source)
+
+    # Generate seccompiler subcommand
+    generate_syscall_parser = subparsers.add_parser(
+        "generate_seccompiler",
+        help="Generate syscall for `rust-vmm/seccompiler` from prepared kernel headers",
+    )
+    default_seccompiler_syscall_path_prefix = f"{os.getcwd()}/{SECCOMPILER_SYSCALL_DIR}"
+    generate_syscall_parser.add_argument(
+        "--output_path",
+        default=default_seccompiler_syscall_path_prefix,
+        help=f"Output directory path (default: {default_seccompiler_syscall_path_prefix})",
+    )
+    generate_syscall_parser.set_defaults(func=generate_seccompiler)
+
+    # Generate kvm-bindings subcommand
+    generate_kvm_bindings_parser = subparsers.add_parser(
+        "generate_kvm_bindings",
+        help="Generate bindings for `rust-vmm/kvm/kvm-bindings` from prepared kernel headers",
+    )
+    default_kvm_bindings_path_prefix = f"{os.getcwd()}/{KVM_BINDINGS_DIR}"
+    generate_kvm_bindings_parser.add_argument(
+        "--output_path",
+        default=default_kvm_bindings_path_prefix,
+        help=f"Output directory path (default: {default_kvm_bindings_path_prefix})",
+    )
+    generate_kvm_bindings_parser.add_argument(
+        "--attribute",
+        help=f"Custom attribute to be added for structures",
+    )
+    generate_kvm_bindings_parser.set_defaults(func=generate_kvm_bindings)
+
+    args = parser.parse_args()
+    args.func(args)
+
+
+if __name__ == "__main__":
+    main()


rustvmm_gen/lib_kernel_source.py
```diff
@@ -0,0 +1,136 @@
...
+    temp_dir = os.path.join("/tmp", f"linux-{version}-prepare")
+
+    if os.path.exists(temp_dir):
+        shutil.rmtree(temp_dir)
+
+    try:
+        os.makedirs(temp_dir)
+        return temp_dir
+    except OSError as e:
+        raise RuntimeError(f"Failed to create temp directory: {e}") from e
```
roypat, line 60: Let's use something that relies on `mkdtemp(2)` here to avoid race conditions and the "delete if exists" dance: https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

rustvmm_gen/lib_kernel_source.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def check_kernel_version(version):
+    """
+    Validate if the input kernel version exists in remote. Supports both X.Y
+    (namely X.Y.0 and .0 should be omitted) and X.Y.Z formats
+    """
+    # Validate version format
+    if not re.match(r"^\d+\.\d+(\.\d+)?$", version):
+        return (False, "Invalid version format. Use X.Y or X.Y.Z")
```
roypat, line 22: Why this tuple of string and bool instead of just raising exceptions for errors, and returning void?

rustvmm_gen/lib_kernel_source.py
```diff
@@ -0,0 +1,136 @@
...
+                temp_dir, f"{os.path.basename(tarball_path).split('.tar')[0]}"
+            )
+            print(f"Extracted to {extract_path}")
+            return extract_path
+    except (tarfile.TarError, IOError) as e:
+        raise RuntimeError(f"Extraction failed: {e}") from e
+
+
+def install_headers(src_dir, arch, install_path, cleanup=False):
+    os.chdir(src_dir)
```
roypat, line 103: This never chdirs back. Maybe `with contextlib.chdir(src_dir):`?

rustvmm_gen/lib_syscall.py
roypat, line 1: This script is specific to rust-vmm/seccompiler, right? If so it should probably live in that repository

rustvmm_gen/lib_kernel_source.py
```diff
@@ -0,0 +1,136 @@
...
+            stderr=subprocess.STDOUT,
+            text=True,
+        )
+        print(result.stdout)
+
+        if cleanup:
+            print(f"Cleaning temporary files at {parent_dir}")
+            shutil.rmtree(parent_dir)
+
+        return True
```
roypat, line 132: I'd also just drop the boolean return value here and simply raise an exception in the failure case

rustvmm_gen/lib_kernel_source.py
```diff
@@ -0,0 +1,136 @@
...
+                "headers_install",
+                f"ARCH={arch}",
+                f"INSTALL_HDR_PATH={abs_install_path}",
+            ],
+            check=True,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            text=True,
+        )
+        print(result.stdout)
```
roypat, line 126: If we just print result.stdout anyway, why not `stdout=subprocess.PIPE` above?

rustvmm_gen/rustvmm_gen.py
```diff
@@ -0,0 +1,116 @@
...
+
+        if not success:
+            raise RuntimeError("Installation process failed")
+
+        print(f"\nSuccessfully installed kernel headers to {args.install_path}")
+        return src_dir
+
+    except Exception as e:
+        print(f"\nError: {str(e)}")
+        exit(1)
```
roypat, line 55: Not really a need for this try-catch + exit, I think. The call to this function is the final statement in `main()`, so we hit an exception we'd exit with non-zero code anyway (and additionally printing "Error:" doesn't really add much imo), so can just drop this.

roypat (commented):
Cool idea! I like having some centralized tools for dealing with all the bindings generation for our various crates. I do think the seccompiler-specific part should go into rust-vmm/seccompiler, however. What I'd find really useful would be to have some wrapper around `bindgen-rs` that can be used to generate the bindings in crates like linux-loader and kvm-bindings (and I maybe xen-sys?) as well, built on top of the kernel header fetching you included here. Do you think that's a possible extension? :o

As a high level comment, maybe let's rename `rustvmm_gen/` to just be `scripts/`, or `tools/`, and drop the `lib_` prefix from the library files, instead moving them to some `lib` subdirectory.

RuoqingHe:
Sorry I forget to mark this as a draft, `kvm-bindings` generation is on its way 😂

Thanks @roypat for reviewing! ❤️

rustvmm_gen/lib_kernel_source.py
RuoqingHe: Removed, changed to `make -C` instead

rustvmm_gen/lib_syscall.py
RuoqingHe: Yes, that lib file is specific to `seccompiler`, and I'm going to write according lib file for `kvm-bindings` and `linux-loader`, all of them are going to need installed headers to continue, WDYT 🤔

roypat: The generic parts, like downloading linux sources and installing headers, can live in rust-vmm-ci, and then there can be scripts in kvm-bindings/linux-leader/seccompiler that each call into the generic rust-vmm-ci scripts. 

For kvm-bindings and linux-loader in particular there's probably some more common parts that can be moved into this repository (I was hoping some script that takes some .h files as input and produces .rs bindings from them?)

RuoqingHe: If we have the required headers, for `kvm-bindings` and `linux-loader` we are only one step away which is a invocation of `bindgen`, and modify something (like for `serde` support), IMO that's similar to `seccompiler` while `seccompiler` don't need `bindgen`

PR #181, opened 2025-03-17
coverage: make crate_features key optional

roypat:
Currently, if a crate wishes to run the coverage test with some cargo features enabled, it must explicitly list them out in the `crate_features` key of the coverage json files.

Change this so that if this key is omitted from the json file, then test_coverage.py passes `--all-features` to cargo.

No functional change, as currently `crate_features` is a mandatory key, so to opt-in to this new behavior, crates must remove this key from their coverage json file(s).

(even with the switch to codecov.io, I suspect we'll still keep these files around, as we'll still need a way to specify exclusions / feature sets, so I don't think this is throwaway work)

### Summary of the PR

### Requirements

requirements:

  test.


stefano-garzarella:
LGTM, but should we document this behavior on the main readme or in https://github.com/rust-vmm/rust-vmm-ci/blob/main/coverage_config_x86_64.json.sample ?

roypat:
> LGTM, but should we document this behavior on the main readme or in [`main`/coverage_config_x86_64.json.sample](https://github.com/rust-vmm/rust-vmm-ci/blob/main/coverage_config_x86_64.json.sample) ?

Ahh, yes, good catch. The README even still talked about `cargo kcov`, which we havent been using in ages haha

stefano-garzarella (approved):

ShadowCurse (approved):

PR #194, opened 2025-10-01
small fixups to cargo-all-features

roypat:
- Turns out any cargo subcommand can be ran like this, instead of just the ones documented on the cargo-all-features crates.io page, so run clippy like this as well
- convert one last forgotten step (unittests-gnu)
- revert riscv64 tests back to using --all-features, as compiling multiple permutations is simply too slow in the emulated environment :(

### Summary of the PR

### Requirements

requirements:

  test.

### DIFF
--- a/.buildkite/test_description.json
+++ b/.buildkite/test_description.json
@@ -5,7 +5,13 @@
       "command": "RUSTFLAGS=\"-D warnings\" cargo all-features build --release --workspace --all-targets",
       "platform": [
         "x86_64",
-        "aarch64",
+        "aarch64"
+      ]
+    },
+    {
+      "test_name": "build-gnu",
+      "command": "RUSTFLAGS=\"-D warnings\" cargo build --all-features --release --workspace --all-targets",
+      "platform": [
         "riscv64"
       ]
     },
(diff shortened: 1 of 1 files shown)


ShadowCurse (dismissed):

roypat:
> @roypat LGTM, but my question now is, should we use `cargo-all-features` everywhere, or for example just with `clippy` and `test` ?

Mh, yeah, I suppose if we do clippy, then we don't strictly need it for build as well, but then, what would the default build test be (because I wouldn't be completely comfortable to not have one at all)? Because if it falls back to --all-features, it won't work for vhost.

> Can we just do some steps (e.g. `build`) with just a subset instead of all combinations?

I don't think so sadly :(

stefano-garzarella:
> > > > @roypat LGTM, but my question now is, should we use `cargo-all-features` everywhere, or for example just with `clippy` and `test` ?
> > > 
> > > 
> > > Mh, yeah, I suppose if we do clippy, then we don't strictly need it for build as well, but then, what would the default build test be (because I wouldn't be completely comfortable to not have one at all)? Because if it falls back to --all-features, it won't work for vhost.
> > 
> > 
> > Default features?
> 
> Mh, yeah, I could be convinced of that. Do you feel strongly about changing this, or are you also happy to leave things as they are?

Not strong as I mentioned :-)

> 
> > > > Can we just do some steps (e.g. `build`) with just a subset instead of all combinations?
> > > 
> > > 
> > > I don't think so sadly :(
> > 
> > 
> > I don't have a strong opinion on this, just an idea to avoid consuming too much CI time.
> 
> I think this is fine, on vm-memory, which has a ton of features, the runtime of doing all the permutation steps is still quicker than build test on riscv without permutations, so generally I don't expect CI time to increase. For crates like vhost that don't support riscv, the build of all permutations takes 5 minutes on my laptop, which is indeed an increase, but hopefully fine? :o

Yep, let's see how it will go, we can eventually implement this later.

PR #195, opened 2025-10-02
ci: add --cfg for cargo doc to disable compile_error!()

roypat:
Some crates, such as vhost, use compile_error!() to indicate that mutually exclusive feature have been specified. This leads to problems when running cargo doc --all-features, which will inevitably trigger these conditions and then lead to documentation not building (for this reason, the vhost documentation currently fails to build on docs.rs for example).

Add a custom cfg that gets passed via RUSTFLAGS in the cargo doc CI step on whose absense these compile_error!() macros can be gated, so that they dont cause issues when building documentation in CI

See also rust-vmm/vhost#326

### Summary of the PR

### Requirements

requirements:

  test.

### DIFF
--- a/.buildkite/test_description.json
+++ b/.buildkite/test_description.json
@@ -29,7 +29,7 @@
     },
     {
       "test_name": "doc",
-      "command": "RUSTDOCFLAGS=\"-D warnings\" cargo doc --workspace --all-features --no-deps"
+      "command": "RUSTDOCFLAGS=\"-D warnings\" RUSTFLAGS=\"--cfg RUTSDOC_disable_feature_compat_errors\" cargo doc --workspace --all-features --no-deps"
     },
     {
       "test_name": "unittests-gnu",


.buildkite/test_description.json
```diff
@@ -29,7 +29,7 @@
     },
     {
       "test_name": "doc",
-      "command": "RUSTDOCFLAGS=\"-D warnings\" cargo doc --workspace --all-features --no-deps"
+      "command": "RUSTDOCFLAGS=\"-D warnings\" RUSTFLAGS=\"--cfg RUTSDOC_disable_feature_compat_errors\"cargo doc --workspace --all-features --no-deps"
```
stefano-garzarella, line 32: I guess the CI is failing for a missing space here:
```suggestion
      "command": "RUSTDOCFLAGS=\"-D warnings\" RUSTFLAGS=\"--cfg RUTSDOC_disable_feature_compat_errors\" cargo doc --workspace --all-features --no-deps"
```

PR #208, opened 2026-05-27
monorepo: Extend rust-vmm-ci to support rust-vmm monorepo

RuoqingHe:
### Summary of the PR

- ci: Enable per-crate selective pipeline generation
- docs: Document --monorepo mode

### Requirements

requirements:

  test.

### DIFF
--- a/.buildkite/autogenerate_pipeline.py
+++ b/.buildkite/autogenerate_pipeline.py
@@ -46,6 +46,12 @@ DOCKER_PLUGIN_CONFIG='{
 }'
 TIMEOUTS_MIN='["style": 30]'
 ```
+
+When invoked with `--workspace-selective`, the script emits one step per
+workspace crate tagged with Buildkite's native `if_changed` property, so each
+crate's shared crate-scoped tests only run when files in that crate (or in a
+workspace-global path) change. Tests tagged with `"scope": "workspace"`
+(e.g. `commit-format`) have no `if_changed` and always run.
 """
 
 import yaml
@@ -54,6 +60,8 @@ import os
 import sys
 import pathlib
 import copy
+import glob
+import tomllib
 
 from argparse import ArgumentParser, RawTextHelpFormatter
 from textwrap import dedent
@@ -84,6 +92,13 @@ TIMEOUTS_MIN = os.getenv("TIMEOUTS_MIN")
 DEFAULT_AGENT_TAG_HYPERVISOR = os.getenv("DEFAULT_AGENT_TAG_HYPERVISOR", "kvm")
 
 BUILDKITE_PATH = pathlib.Path(__file__).parent.resolve()
+
+# Per-crate file listing the platforms CI should run on.
+PLATFORMS_FILE = ".platform"
+
+# Paths added to every shared crate-scoped step's `if_changed`, so a change to
+# the workspace as a whole fires each crate's shared tests.
+WORKSPACE_GLOBAL_PATHS = ("Cargo.toml", ".buildkite/**", "rust-vmm-ci/**")
 
 
 class BuildkiteStep:
@@ -232,16 +247,26 @@ class BuildkiteStep:
         conditional = input.get("conditional")
         timeout = input.get("timeout_in_minutes")
         queue = input.get("queue")
+        crate = input.get("crate")
+        crate_path = input.get("crate_path")
 
         # Mandatory keys.
         assert test_name, "Step is missing test name."
         platform_string = f"-{platform}" if platform else ""
-        self.step_config["label"] = f"{test_name}{platform_string}"
+        # When the step is crate-scoped, prefix the label with the crate to disambiguate.
+        crate_string = f"{crate}: " if crate else ""
+        self.step_config["label"] = f"{crate_string}{test_name}{platform_string}"
 
         assert command, "Step is missing command."
         if "{target_platform}" in command:
             assert platform, "Command requires platform, but platform is missing."
             command = command.replace("{target_platform}", platform)
+        if "{crate}" in command:
+            assert crate, "Command requires crate, but crate is missing."
+            command = command.replace("{crate}", crate)
+        if "{crate_path}" in command:
+            assert crate_path, "Command requires crate path, but crate path is missing."
+            command = command.replace("{crate_path}", crate_path)
         # Modify command and tag name for `riscv64` CI
         if platform == "riscv64":
             # Wrap command with '' to avoid escaping early by `ENTRYPOINT`
@@ -282,6 +307,9 @@ class BuildkiteStep:
             "test_name",
             "queue",
             "hypervisor",
+            "crate",
+            "crate_path",
+            "scope",
         ]
         additional_keys = {
             k: v
@@ -305,6 +333,38 @@ class BuildkiteConfig:
     def __init__(self):
         self.bk_config = None
 
+    @staticmethod
+    def _skip_test(test_name):
+        """Return whether `test_name` is excluded via the `TESTS_TO_SKIP` env."""
+        return bool(TESTS_TO_SKIP) and test_name in json.loads(TESTS_TO_SKIP)
+
+    def _append_test_steps(
+        self, test, platform_allowlist, crate=None, crate_path=None, if_changed=None
+    ):
+        """Append the steps for `test`, one per allowed platform, scoped to `crate`."""
+        platforms = test.get("platform")
+        # The platform is optional.
+        if not platforms:
+            platforms = [None]
+
+        for platform in platforms:
+            # Filter test enabled in platform_allowlist
+            if platform is not None and platform not in platform_allowlist:
+                # Skip disabled platform
+                continue
+
+            step_input = copy.deepcopy(test)
+            step_input["platform"] = platform
+            step_input["crate"] = crate
+            step_input["crate_path"] = crate_path
+            if if_changed is not None:
+                step_input["if_changed"] = if_changed
+            if not step_input.get("hypervisor"):
+                step_input["hypervisor"] = DEFAULT_AGENT_TAG_HYPERVISOR
+
+            step = BuildkiteStep()
+            self.bk_config["steps"].append(step.build(step_input))
+
     def build(self, input, platform_allowlist):
         """Build the final Buildkite configuration fron the json input."""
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -191,6 +191,53 @@ script that autogenerates the main pipeline can be used with the option
...
+./rust-vmm-ci/.buildkite/autogenerate_pipeline.py --monorepo -t .buildkite/test_description.json | buildkite-agent pipeline upload
+```
+
+The generator diffs the working tree against the base branch (`main`, or
+`BUILDKITE_PULL_REQUEST_BASE_BRANCH` when set), maps the changed files to their
+crates, adds the workspace dependents of those crates (via `cargo metadata`),
+and emits steps only for the result. Each crate uses its own `.platform` and
+`coverage_config_ARCH.json`, and the coverage test runs from the crate's
+directory with `--test-scope=crate`. Changing a workspace-global path (the root
+`Cargo.toml`, `Cargo.lock`, `.buildkite`, or the `rust-vmm-ci` submodule) runs
```
stefano-garzarella: But `Cargo.lock` is not part of `WORKSPACE_GLOBAL_PATHS`.

I'm not asking to add it or not, just pointing out there is a mismatch.

.buildkite/autogenerate_pipeline.py
```diff
@@ -305,6 +330,33 @@ class BuildkiteConfig:
     def __init__(self):
         self.bk_config = None
 
+    def _skip_test(self, test_name):
+        """Return whether `test_name` is excluded via the `TESTS_TO_SKIP` env."""
+        return bool(TESTS_TO_SKIP) and test_name in json.loads(TESTS_TO_SKIP)
```
stefano-garzarella, line 339: Is the json cached in some way or are we re-loading it on every call?

README.md
```diff
@@ -191,6 +191,53 @@ script that autogenerates the main pipeline can be used with the option
...
+
+```bash
+./rust-vmm-ci/.buildkite/autogenerate_pipeline.py --monorepo -t .buildkite/test_description.json | buildkite-agent pipeline upload
+```
+
+The generator diffs the working tree against the base branch (`main`, or
+`BUILDKITE_PULL_REQUEST_BASE_BRANCH` when set), maps the changed files to their
+crates, adds the workspace dependents of those crates (via `cargo metadata`),
+and emits steps only for the result. Each crate uses its own `.platform` and
+`coverage_config_ARCH.json`, and the coverage test runs from the crate's
```
stefano-garzarella, line 218: Are we supporting multiples coverage files with this PR or this is a future work?

.buildkite/autogenerate_pipeline.py
```diff
@@ -408,6 +587,32 @@ def generate_pipeline(config_file, platform_allowlist):
         ),
         default=f"{os.getcwd()}/.platform",
     )
+    parser.add_argument(
+        "--monorepo",
```
stefano-garzarella: `--monorepo` seems more related to a case where use this feature, but in theory we can also use in other workspaces. What about a different name for this parameter?

WDYT about `--workspace-selective` or something like that? in order to refer more to the feature

README.md
```diff
@@ -191,6 +191,53 @@ script that autogenerates the main pipeline can be used with the option
...
+The `rust-vmm/rust-vmm` monorepo hosts several crates in a single Cargo
+workspace and can run CI selectively, so that a change to one crate does not
+retest the whole workspace. Passing `--monorepo` enables this:
+
+```bash
+./rust-vmm-ci/.buildkite/autogenerate_pipeline.py --monorepo -t .buildkite/test_description.json | buildkite-agent pipeline upload
+```
+
+The generator diffs the working tree against the base branch (`main`, or
+`BUILDKITE_PULL_REQUEST_BASE_BRANCH` when set), maps the changed files to their
```
stefano-garzarella: Should we mention also the `--base-branch` parameter ?
Also, should we mention in the first commit that the commit is adding also that parameter and why?

stefano-garzarella (commented):
@RuoqingHe many thanks for this!! I left some comments, but in general LGTM.

README.md
RuoqingHe: Since we are shipping libraries, I eventually decide to remove Cargo.lock, good catch

README.md
RuoqingHe: Noted

.buildkite/autogenerate_pipeline.py
RuoqingHe: IIUC it is parsing on every call, I kept the original behaviour

README.md
RuoqingHe: IIUC it's supported already, just because we wrote x86_64 in test_description.json so now we only have x86_64 coverage

andreeaflorescu (changes requested):
This is up for debate, but I would rather use the support that Buildkite already has for mono-repo, and build our features around that instead.

Here is the documentation: https://buildkite.com/docs/pipelines/best-practices/working-with-monorepos.

I also think is important to explicitly define what we want to achieve with this change:
- each crate can define its own supported platforms
- we can run custom tests for each crate
- we can run a minimal set of tests across all crates to check that changes to one crate don't break compatibility, without overwhelming the CI
This part should be defined in the readme.

I'll add more comments in the code with the specific things that I think we can improve.

.buildkite/autogenerate_pipeline.py
```diff
@@ -85,6 +91,12 @@
 
 BUILDKITE_PATH = pathlib.Path(__file__).parent.resolve()
 
+# Per-crate file listing the platforms CI should run on.
+PLATFORMS_FILE = ".platform"
+
+# Touching any of these top-level paths triggers CI for every crate.
```
andreeaflorescu: Does that mean that we want to trigger the custom, per-crate pipelines as well?

.buildkite/autogenerate_pipeline.py
```diff
@@ -357,6 +417,112 @@ def determine_allowlist(config_file):
...
+        )
+        if owner is None:
+            # A file outside every crate is treated as global.
+            return set(dirs)
+        changed_crates.add(owner)
+
+    return changed_crates
+
+
+def reverse_dependency_closure(changed_crates, deps):
```
andreeaflorescu: I find this a bit more complicated than it needs to be, making it error prone. If we change something in vmm-sys-util, and we run `cargo build` and `cargo test` at the workspace level, then we should catch most of the issues that might come from the change.

With this simple approach we can get rid of most of the code here, while still catching 99% of the bugs.

If we find that this is not enough, then we can add a more complex way of deciding which tests to run. This way we don't optimize early for a problem that we didn't run into.

.buildkite/autogenerate_pipeline.py
```diff
@@ -357,6 +417,112 @@ def determine_allowlist(config_file):
...
+        dirs[package["name"]] = os.path.relpath(manifest_dir, root)
+        raw_deps[package["name"]] = {dep["name"] for dep in package["dependencies"]}
+
+    # Keep intra-workspace edges only.
+    members = set(dirs)
+    deps = {name: raw_deps[name] & members for name in members}
+    return dirs, deps
+
+
+def changed_files(base_branch):
```
andreeaflorescu: We don't need to add our own logic here. We should just use `if_changed` property from Buildkite.

andreeaflorescu:
I would propose we start simple here, and iterate on the solution as needed.

## What we MUST support
 We need to support running the same tests with the mono-repo that we were able to run when having separate repositories. This is the no 1 priority because otherwise we might introduce bugs. This means two things:
- support for running custom pipelines defined per crate
- support for running the tests on multiple platforms (as defined in the `.platform` per crate).

Let's focus on these two things, and use the support from Buildkite to achieve our goals and minimize the required changes. Code that we don't write, we don't need to maintain and this is going to save us in the long run.

## Possible Approaches

### Option 1: Run Tests at Crate Level
This would mean that we need to identify which tests are path dependent. You are already doing this work in your PR. Global tests are things like commit formatting, and path dependent ones are `cargo test`. All the cargo related commands should have only one change, which is specify `-p <package>`. Instead of having a single pipeline, we can have 2. One is path independent, and always runs once. The other ones are path dependent, and run once for each `if_changed` path. The path-dependent ones are also typically platform dependent. So we can read the `.platform` file in the crate to decide which tests we need to generate per platform.

Advantages:
- Clean separation of what we need to run when. This makes tests easier to track.
- Allows tests to run on the platforms that they were intended to with a single configuration.

Disadvantages:
- We need a global pipeline to catch issues generated by a change in a dependent crate. NOTE: I don't think this is an issue right now. Before the mono-repo we would also not catch these issues, so we can just treat it as an improvement that we do once we have on-par CI for mono-repo as we did for individual crates.

### Option 2: Run All Tests at Workspace Level

With this approach we just run all tests at the workspace level. At the workspace level we have a `.platform` file that has the common denominator of platforms supported by all crates. We still need to run the tests specifying the package for all the platforms that are not at the workspace level.

Disadvantages:
-  We run more tests than would be strictly necessary. If the CI times increase significantly because of this, we optimize it later on.
- We still need to run per-crate tests for platforms that are specific to the crate. The logic might become more complicated then with Option 1.

Advantage: 
- We automatically catch any cross-dependency issues introduced. If a change in vmm-sys-util breaks something in vm-memory we know by the virtue of running all tests at the workspace level. This advantage is weak though because we can just have a separate global pipeline that runs `cargo test` and `cargo build` at the workspace level instead.

Irrespective of the approach we take, we need to add support for running custom pipelines if they exist in the code being changed. To be able to do that, we can extend the pipeline generation to add support for `if_changed`. We already support passing the path of the pipeline, so no other changes should be required.


My proposal is to go with Option 1. I believe is the least disruptive one, and the one that helps us the most in the lon run. WDYT?

.buildkite/autogenerate_pipeline.py
RuoqingHe: Yes, I think changes in the repo root should trigger CI for every crate

.buildkite/autogenerate_pipeline.py
RuoqingHe: Removed

.buildkite/autogenerate_pipeline.py
RuoqingHe: Thanks for sharing this, I seldom use buildkite, updated with `if_changed`

stefano-garzarella (dismissed):
LGTM! Thanks for that!

.buildkite/autogenerate_pipeline.py
```diff
@@ -305,6 +332,37 @@ class BuildkiteConfig:
     def __init__(self):
         self.bk_config = None
 
+    def _skip_test(self, test_name):
```
andreeaflorescu: `self` is not used, you should remove it.

andreeaflorescu:
The next steps for this to be merged are:
1. Create a test pipeline in Buildkite and prove how it's going to work. This should include 2 steps: the generic pipeline, and calling into the custom vm-memory pipeline. We need to check that tests for the platforms that vm-memory support are properly generated.
2. Update the tests that should only be run as part of the workspace.

.buildkite/autogenerate_pipeline.py
```diff
@@ -357,6 +440,29 @@ def determine_allowlist(config_file):
...
+        return []
+    with open(path) as json_file:
+        return json.load(json_file).get("tests", [])
+
+
+def workspace_members():
+    """Return `{crate_name: relative_path}` for each workspace member."""
+    metadata = json.loads(
+        subprocess.check_output(
+            ["cargo", "metadata", "--no-deps", "--format-version", "1"]
```
andreeaflorescu: We cannot run `cargo` when autogenerating the pipeline because this script runs on a general purpose machine, that doesn't have anything rust related installed on it. Can you keep the script agnostic so it can still run on any machine?

.buildkite/autogenerate_pipeline.py
```diff
@@ -408,6 +527,21 @@ def generate_pipeline(config_file, platform_allowlist):
         ),
         default=f"{os.getcwd()}/.platform",
     )
+    parser.add_argument(
+        "--workspace-selective",
+        action="store_true",
```
andreeaflorescu: Can we make this an optional parameter whose default value is False instead? Because of a limitation on the infrastructure side of things we currently don't support options, only parameters. The infrastructure code is a bit of a mess right now, and would require a complete re-write to make it more future proof. We don't really have the time right now to invest in this effort unfortunately.

.buildkite/autogenerate_pipeline.py
RuoqingHe: Using python tomllib to extract the members

.buildkite/autogenerate_pipeline.py
```diff
@@ -357,6 +442,43 @@ def determine_allowlist(config_file):
...
+        return json.load(json_file).get("tests", [])
+
+
+def workspace_members():
+    """Return `{crate_name: relative_path}` for each workspace member.
+
+    Parses the workspace's Cargo.toml manifests with the stdlib `tomllib`
+    module so this script does not require `cargo` on the host.
+    """
+    with open("Cargo.toml", "rb") as f:
```
andreeaflorescu, line 460: Why do we need to open with `rb`? Why does `r` not work? The `Cargo.toml` file is not a binary.

.buildkite/autogenerate_pipeline.py
```diff
@@ -408,6 +543,23 @@ def generate_pipeline(config_file, platform_allowlist):
         ),
         default=f"{os.getcwd()}/.platform",
     )
+    parser.add_argument(
+        "--workspace-selective",
+        type=lambda v: str(v).lower() == "true",
```
andreeaflorescu: We can update the invoke to write True. No need for this lambda.

.buildkite/autogenerate_pipeline.py
```diff
@@ -408,6 +543,23 @@ def generate_pipeline(config_file, platform_allowlist):
         ),
         default=f"{os.getcwd()}/.platform",
     )
+    parser.add_argument(
```
andreeaflorescu, line 549: Can you please add a comment here to specify that we have this as an argument and not an option because of a limitation on the infrastructure side which allows only arguments (no options) for parameters passed to this script?

andreeaflorescu: Otherwise in the future people might tried to update it and it would fail.

README.md
```diff
@@ -191,6 +191,62 @@ script that autogenerates the main pipeline can be used with the option
...
 ```
 
+### Workspace-Selective Pipeline
+
+The `rust-vmm/rust-vmm` monorepo hosts several crates in a single Cargo
+workspace and can run CI selectively, so that a change to one crate does not
+retest the whole workspace. Passing `--workspace-selective true` enables this:
+
+```bash
+./rust-vmm-ci/.buildkite/autogenerate_pipeline.py --workspace-selective true -t .buildkite/test_description.json | buildkite-agent pipeline upload
```
andreeaflorescu: This command will need updating once we pass True and not `true`.

.buildkite/autogenerate_pipeline.py
RuoqingHe: `tomllib` documentation mandates binary mode 🤷 I didn't dive into impl so not sure why

.buildkite/autogenerate_pipeline.py
RuoqingHe: Dropped 👍

andreeaflorescu (approved):

stefano-garzarella (approved):

PR #210, opened 2026-07-09
allow tests to define if_changed and crate_path

andreeaflorescu:
### Summary of the PR

We now check whether the test defines if_changed and crate_path before using the ones from the workspace. If the tests does define these two, then we don't overwrite them with the default workspace configuration.

This is needed so that custom pipelines for tests can specify as if changed the crate that they're intended to. Without this change we end up running the custom tests for all crates that are changed instead of running it just for the crate that they're intended.

### Requirements

requirements:

  test.


stefano-garzarella (approved):

RuoqingHe (approved):
Looks reasonable to me, thanks @andreeaflorescu ❤️

epilys (approved):

